PR #3, opened 2025-07-06
feat: parsec-vdd

galaxy-s10:


example/lib/main.dart
```diff
@@ -243,6 +243,54 @@ class _SimulatorScreenState extends State<SimulatorScreen> {
     HardwareSimulator.performTouchEvent(x, y, touchId, false, 0);
   }
 
+  int _currentDisplayId = -1; // 存储当前创建的显示器ID
+
+  void createDisplay() async {
+    try {
+      // First initialize Parsec
+      bool initialized = await HardwareSimulator.initializeParsec();
```
zhuhaichao518, line 251: Please use some name like initializeParsecVdd, we are not using the parsec project but parsec-vdd.

PR #13, opened 2026-05-14
fix(android): decouple locked key bridge from gamepads

zhuhaichao518:
## Summary

- Add a hardware-simulator-owned Android activity bridge for locked key events.
- Remove the Android plugin's direct dependency on `gamepads_android`.
- Keep locked mouse keyboard forwarding in `hardware_simulator` instead of extending the gamepads activity interface.

## Why

The Android implementation was calling `registerLockedKeyEventHandler` on `GamepadsCompatibleActivity`, but the current `cloudplayplus_gamepads` Android interface does not expose that method. This caused `:hardware_simulator:compileDebugKotlin` to fail in the main CloudPlayPlus app.

## Validation

- Verified through the main app with `flutter build apk --debug` after wiring the bridge in `cloudplayplus`.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -66,7 +66,3 @@ android {
         }
     }
 }
-
-dependencies {
-    implementation project(':gamepads_android')
-}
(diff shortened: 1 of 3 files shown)


zhuhaichao518:
@claude review

PR #14, opened 2026-05-14
fix(macos): preserve drag metadata for injected mouse input

zhuhaichao518:
## Summary

- Route macOS synthetic mouse movement and button events through a Sunshine-style `postMouse` helper.
- Track injected mouse button state locally so remote drag movement emits dragged event types.
- Populate mouse button number, click state, movement deltas, and warp the cursor after posting synthetic events.

## Root Cause

Some macOS games do not treat remote mouse movement as a real drag when the synthetic event only changes type/location. Sunshine fixed a similar in-game mouse issue by carrying drag state and filling the CGEvent mouse metadata, especially delta fields.

## Validation

- User tested macOS host drag in Slay the Spire successfully.
- `flutter build macos --debug --no-pub` passed from the parent Flutter app checkout.


zhuhaichao518:
@claude review

PR #15, opened 2026-05-18
fix(ios): find embedded Flutter view for cursor lock

zhuhaichao518:
## Summary
- make iOS cursor-lock controller lookup recurse through child/presented view controllers
- support CloudPlayPlus's NativeOverlayViewController root while keeping the existing lockCursor behavior

## Validation
- from parent app: flutter analyze
- from parent app: LANG=en_US.UTF-8 LC_ALL=en_US.UTF-8 flutter build ios --debug --no-codesign

## Notes
- This PR is required by the parent cloudplayplus PR so iOS lockCursor reaches the embedded FlutterViewController.


zhuhaichao518:
@claude review

PR #16, opened 2026-05-19
fix(linux): use inputtino for mouse simulation

zhuhaichao518:
## Summary
- add inputtino as a Linux submodule dependency for uinput-backed virtual mouse events
- route Linux mouse move/click/scroll method-channel calls through a persistent inputtino mouse
- keep keyboard on the existing X11/XTest path and document Linux backend requirements

## Validation
- flutter build linux --debug (plugins/hardware_simulator/example)
- ./build/linux/x64/debug/plugins/hardware_simulator/hardware_simulator_test
- flutter test (plugins/hardware_simulator)
- flutter build linux --debug (main cloudplayplus app)

No UI change; release preview build not required.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "third_party/inputtino"]
+	path = third_party/inputtino
+	url = https://github.com/games-on-whales/inputtino.git
--- a/README.md
+++ b/README.md
@@ -2,6 +2,6 @@
 
 This plugin simulates mouse & keyboard input. Currently it is not well documented. 
 
-Current it supports to simulate: mouse & keyboard for Windows/MacOS. XBOX Game Controller for windows. See the example for details.
+Current it supports to simulate: mouse & keyboard for Windows/macOS/Linux (Linux mouse uses uinput via inputtino; Linux keyboard still requires X11 + XTest). XBOX Game Controller for windows. See the example for details.
 
 Any pull request is welcome. It is designed for https://github.com/zhuhaichao518/cloudplayplus_stone.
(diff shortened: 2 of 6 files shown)


zhuhaichao518:
@claude review

zhuhaichao518:
@claude review

zhuhaichao518:
@claude review

zhuhaichao518:
[P1] linux/CMakeLists.txt:44-47
pkg_check_modules(LIBEVDEV QUIET libevdev) does not create PkgConfig::LIBEVDEV, but line 47 passes PkgConfig::LIBEVDEV into inputtino’s custom library path. On machines with libevdev-dev installed, CMake can fail because that imported target does not exist. Use pkg_check_modules(LIBEVDEV QUIET IMPORTED_TARGET libevdev) or pass the actual ${LIBEVDEV_LIBRARIES}/link dirs.

[P2] linux/hardware_simulator_plugin.cc:466-489
clear_all_pressed_events() returns DISPLAY_UNAVAILABLE before releasing mouse buttons. Since Linux mouse buttons now use uinput and can be pressed without XTest, losing DISPLAY or running mouse-only on Wayland/headless can leave a virtual mouse button stuck down. Release tracked mouse buttons independently of X11, then release keys only if XTest is available.

[P2] linux/hardware_simulator_plugin.cc:364, linux/hardware_simulator_plugin.cc:398, linux/hardware_simulator_plugin.cc:426-429
inputtino switches between relative and absolute internal mouse devices based on the last move call. If a button is pressed after an absolute move, then a relative move occurs before release, the release may be sent to the relative device while the absolute device remains pressed. Track the device/mode used for each pressed button or keep button events on a single mouse device path.

zhuhaichao518:
@claude review

zhuhaichao518:
Let‘s merge first and fix bugs later.

PR #19, opened 2026-06-15
fix(cursor): use endian-stable position payloads

zhuhaichao518:
## Summary
- Encode cursor position callbacks as two normalized little-endian `uint16` values instead of native `float` bytes.
- Clamp non-finite and out-of-range percentages at the platform boundary.
- Update the example decoder to read the 4-byte position payload.

## Rationale
Cursor positions are bounded percentages, so a normalized 16-bit representation is precise enough for screen placement and cuts the callback payload from 8 bytes to 4 bytes. More importantly, the payload no longer depends on native float memory layout or host endianness.

## Validation
- `flutter test`
- `cmake --build . --config Release --target hardware_simulator_plugin` from the main app Windows build directory
- `flutter analyze` was checked; the plugin still has a pre-existing example lint backlog, but this change does not introduce a build break

### DIFF
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -157,16 +157,16 @@ class _SimulatorScreenState extends State<SimulatorScreen> {
         // 获取显示器ID和鼠标位置信息
         int screenId = messageInfo;
 
-        // 检查是否有位置数据（新版本有8字节，旧版本可能为空）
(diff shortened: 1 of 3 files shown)


zhuhaichao518:
@claude review

zhuhaichao518:
@claude review

PR #36, opened 2026-07-23
feat(logging): integrate shared native logger

zhuhaichao518:
## What changed

- link the Windows plugin to the host-provided `cpp_log_plugin` target when available
- route representative ViGEm game-controller diagnostics through the shared native logger
- preserve standalone builds with a stderr fallback when the host does not provide `cpp_log`
- avoid a mandatory Dart path dependency on a sibling package

## Why

Native plugin diagnostics should share CloudPlayPlus's bounded asynchronous logger and rotating file sink instead of maintaining separate output paths, while this plugin must remain independently buildable.

## Validation

- `flutter pub get`
- standalone `example`: `flutter build windows --debug`
- CloudPlayPlus host Windows Release build with the shared `cpp_log.dll`
- `flutter analyze` still reports the repository's existing 268 example lint items; this PR does not change Dart sources

### DIFF
--- a/windows/gamecontroller_manager.cc
+++ b/windows/gamecontroller_manager.cc
@@ -1,5 +1,27 @@
 #include "gamecontroller_manager.h"
 
+#if defined(CPP_LOG_AVAILABLE)
+#include <cpp_log/cpp_log.h>
+#else
+#include <cstdarg>
+#include <cstdio>
+
+namespace {
+void FallbackLog(const char* level, const char* tag, const char* format, ...) {
+  std::fprintf(stderr, "[%s] [%s] ", level, tag);
+  va_list args;
+  va_start(args, format);
+  std::vfprintf(stderr, format, args);
+  va_end(args);
+  std::fputc('\n', stderr);
+}
+}  // namespace
+
+#define CPPLOG_INFO(tag, ...) FallbackLog("INFO", (tag), __VA_ARGS__)
+#define CPPLOG_WARN(tag, ...) FallbackLog("WARN", (tag), __VA_ARGS__)
+#define CPPLOG_ERROR(tag, ...) FallbackLog("ERROR", (tag), __VA_ARGS__)
+#endif
+
 #include <sstream>
 #include <Xinput.h>
 
@@ -10,14 +32,14 @@ std::array<PVIGEM_TARGET, 4> GameControllerManager::controllers = {};
 int GameControllerManager::InitializeVigem() {
   vigem_client = vigem_alloc();
   if (vigem_client == nullptr) {
-    std::cerr << "Uh, not enough memory to do that?!" << std::endl;
+    CPPLOG_ERROR("GAMEPAD", "ViGEm client allocation failed");
     return -1;
   }
 
   const auto retval = vigem_connect(vigem_client);
   if (!VIGEM_SUCCESS(retval)) {
-    std::cerr << "ViGEm Bus connection failed with error code: 0x"
-              << std::hex << retval << std::endl;
+    CPPLOG_ERROR("GAMEPAD", "ViGEm Bus connection failed: 0x%X",
+                 static_cast<unsigned int>(retval));
     return -1;
   }
   initialized = true;
@@ -43,22 +65,22 @@ int GameControllerManager::CreateGameController() {
       // Error handling
       //
       if (!VIGEM_SUCCESS(pir)) {
-        std::cerr << "Target plugin failed with error code: 0x" << std::hex
-                << pir << std::endl;
+        CPPLOG_ERROR("GAMEPAD", "ViGEm target add failed: 0x%X",
+                     static_cast<unsigned int>(pir));
         return -1;
       }
-      std::cout << "GameController created in slot " << i + 1 << std::endl;
+      CPPLOG_INFO("GAMEPAD", "Game controller created in slot %d", i + 1);
       return i + 1;
     }
   }
 
-  std::cerr << "No available slot for GameController!" << std::endl;
+  CPPLOG_WARN("GAMEPAD", "No available game controller slot");
   return -1;
 }
 
 bool GameControllerManager::RemoveGameController(int id) {
   if (id < 1 || id > 4) {
-    std::cerr << "Invalid slot id: " << id << std::endl;
+    CPPLOG_WARN("GAMEPAD", "Invalid game controller slot: %d", id);
     return false;
   }
 
@@ -69,15 +91,15 @@ bool GameControllerManager::RemoveGameController(int id) {
     // Error handling
     //
     if (!VIGEM_SUCCESS(pir)) {
-        std::cerr << "GameController remove failed with error code: 0x" << std::hex
-            << pir << std::endl;
+        CPPLOG_ERROR("GAMEPAD", "Game controller removal failed: 0x%X",
+                     static_cast<unsigned int>(pir));
         return false;
     }
-    std::cout << "GameController removed from slot " << id << std::endl;
+    CPPLOG_INFO("GAMEPAD", "Game controller removed from slot %d", id);
     return true;
   }
 
-  std::cerr << "Slot " << id << " is already empty!" << std::endl;
+  CPPLOG_WARN("GAMEPAD", "Game controller slot %d is already empty", id);
   return false;
 }
 
@@ -103,10 +125,9 @@ bool GameControllerManager::DoControllerAction(int id, std::string& action) {
         *reinterpret_cast<XUSB_REPORT*>(&gamepad));
 
     if (!VIGEM_SUCCESS(pir)) {
-        std::cerr << "GameController remove failed with error code: 0x" << std::hex
-            << pir << std::endl;
+        CPPLOG_ERROR("GAMEPAD", "Game controller update failed: 0x%X",
+                     static_cast<unsigned int>(pir));
         return false;
     }
-    std::cout << "GameController removed from slot " << id << std::endl;
     return true;
 }
--- a/windows/gamecontroller_manager.cc
+++ b/windows/gamecontroller_manager.cc
@@ -1,26 +1,5 @@
 #include "gamecontroller_manager.h"
-
-#if defined(CPP_LOG_AVAILABLE)
-#include <cpp_log/cpp_log.h>
-#else
-#include <cstdarg>
-#include <cstdio>
-
-namespace {
-void FallbackLog(const char* level, const char* tag, const char* format, ...) {
-  std::fprintf(stderr, "[%s] [%s] ", level, tag);
-  va_list args;
-  va_start(args, format);
-  std::vfprintf(stderr, format, args);
-  va_end(args);
-  std::fputc('\n', stderr);
-}
-}  // namespace
-
-#define CPPLOG_INFO(tag, ...) FallbackLog("INFO", (tag), __VA_ARGS__)
-#define CPPLOG_WARN(tag, ...) FallbackLog("WARN", (tag), __VA_ARGS__)
-#define CPPLOG_ERROR(tag, ...) FallbackLog("ERROR", (tag), __VA_ARGS__)
-#endif
+#include "cpp_log_shim.h"
 
 #include <sstream>
 #include <Xinput.h>
--- a/windows/CMakeLists.txt
+++ b/windows/CMakeLists.txt
@@ -24,6 +24,7 @@ list(APPEND PLUGIN_SOURCES
   "hardware_simulator_plugin.h"
   "cursor_monitor.cc"
   "cursor_monitor.h"
+  "cpp_log_shim.h"
   "gamecontroller_manager.cc"
   "gamecontroller_manager.h"
   "notification_window.cc"
--- a/windows/CMakeLists.txt
+++ b/windows/CMakeLists.txt
@@ -59,6 +59,13 @@ apply_standard_settings(${PLUGIN_NAME})
 set_target_properties(${PLUGIN_NAME} PROPERTIES
   CXX_VISIBILITY_PRESET hidden)
 target_compile_definitions(${PLUGIN_NAME} PRIVATE FLUTTER_PLUGIN_IMPL)
+if(TARGET cpp_log_plugin)
+  target_compile_definitions(${PLUGIN_NAME} PRIVATE CPP_LOG_AVAILABLE=1)
+  target_link_libraries(${PLUGIN_NAME} PRIVATE cpp_log_plugin)
+else()
+  message(STATUS
+    "hardware_simulator: cpp_log target unavailable; native logging disabled")
+endif()
 
 # Source include directories and library dependencies. Add any plugin-specific
 # dependencies here.
--- a/windows/cpp_log_shim.h
+++ b/windows/cpp_log_shim.h
@@ -1,0 +1,47 @@
+#ifndef HARDWARE_SIMULATOR_WINDOWS_CPP_LOG_SHIM_H_
+#define HARDWARE_SIMULATOR_WINDOWS_CPP_LOG_SHIM_H_
+
+#if defined(CPP_LOG_AVAILABLE)
+#include <cpp_log/cpp_log.h>
+#else
+#include <cstdarg>
+#include <cstdio>
+
+namespace hardware_simulator {
+namespace logging {
+
+inline void FallbackLog(const char* level, const char* tag,
+                        const char* format, ...) {
+  char message[2048] = {};
+  va_list args;
+  va_start(args, format);
+  std::vsnprintf(message, sizeof(message), format, args);
+  va_end(args);
+  std::fprintf(stderr, "[%s] [%s] %s\n", level,
+               tag && *tag ? tag : "NATIVE", message);
+}
+
+}  // namespace logging
+}  // namespace hardware_simulator
+
+#define CPPLOG_TRACE(tag, ...)                                      \
+  ::hardware_simulator::logging::FallbackLog("TRACE", (tag),        \
+                                              __VA_ARGS__)
+#define CPPLOG_DEBUG(tag, ...)                                      \
+  ::hardware_simulator::logging::FallbackLog("DEBUG", (tag),        \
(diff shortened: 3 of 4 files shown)


zhuhaichao518:
@hxyshadow review

windows/gamecontroller_manager.cc
```diff
@@ -1,5 +1,27 @@
 #include "gamecontroller_manager.h"
 
+#if defined(CPP_LOG_AVAILABLE)
```
hxyshadow: 💡 **Suggestion (forward-looking DRY):** This 20-line `CPP_LOG_AVAILABLE` fallback shim (`FallbackLog` + the three `CPPLOG_*` macros) is TU-local to this file. The PR body scopes this to *representative* diagnostics, and the other Windows sources still emit raw output — `hardware_simulator_plugin.cpp`, `SmartKeyboardBlocker.cpp`, `virtual_display.cc`, `virtual_display_control.cc` all still use `std::cout`/`OutputDebugString`/`printf`. When you migrate those in a follow-up, each will need this identical block copy-pasted. Consider extracting the shim now into a small shared internal header (e.g. `windows/cpp_log_shim.h`) so the next N files just `#include` it and stay consistent — avoids the shim drifting per-file.

windows/gamecontroller_manager.cc
```diff
@@ -1,5 +1,27 @@
...
+#include <cpp_log/cpp_log.h>
+#else
+#include <cstdarg>
+#include <cstdio>
+
+namespace {
+void FallbackLog(const char* level, const char* tag, const char* format, ...) {
+  std::fprintf(stderr, "[%s] [%s] ", level, tag);
```
hxyshadow: 💡 **Suggestion (fallback atomicity):** The fallback emits the `[LEVEL] [TAG] ` prefix and the formatted message as two separate `stderr` writes. `fprintf`/`vfprintf` are each individually thread-safe on MSVCRT, but the two calls are not atomic together, so concurrent gamepad calls (platform-channel thread) can interleave a prefix with another thread's message. Cheap fix: `vsnprintf` into a stack buffer once, then a single `fprintf(stderr, "[%s] [%s] %s\n", level, tag, buf)`. Only matters on the standalone/example build (the host `cpp_log` path already serializes via its ring buffer), so non-blocking.

hxyshadow (changes requested):
**🔴 REQUEST_CHANGES(需修改)**

## Code Review: #36 — feat(logging): integrate shared native logger

**结论：Approve ✅**（0 blocking，2 条非阻塞 💡）
**Reviewed against:** `76bcf8e93acfc491f66d9fb7b12c11440735a55b`

首轮 review（无前序 reviewer，作者直接 `@hxyshadow review`）。改动小而干净：把 ViGEm 游戏手柄的诊断输出从 `std::cout`/`std::cerr` 迁移到共享 native logger `cpp_log`，并保留独立可编译能力（host 未提供 `cpp_log` 时用 stderr fallback）。

### ✅ 跨 PR 契约校验（重点）
这是一个跨仓库协同集成 PR，真正的契约方在兄弟 PR 里，本仓库 diff 只是消费端。我对着协同分支逐项核对，契约**成立**：

- **target 名**：`windows/CMakeLists.txt` 用 `if(TARGET cpp_log_plugin)` 门控。协同的 `CloudPlayPlus/cloudplayplus_log#2`（host PR `cloudplayplus#407` 的依赖 PR）把 `PLUGIN_NAME` 设为 `"cpp_log_plugin"` — 名字对得上 ✅。（注意：cpp_log **旧** commit 里 target 叫 `cpp_log`、且无 public header；只有 spdlog 重构后的 `cpp_log_log#2` 才改名并导出 header。本仓库必须与该 PR 一起合并才成立。）
- **头文件**：`#include <cpp_log/cpp_log.h>`。`cloudplayplus_log#2` 新增 `include/cpp_log/cpp_log.h`，且 `windows/CMakeLists.txt` 用 `target_include_directories(... PUBLIC ../include ...)` 导出 — 消费端能解析到 ✅。（旧版是 `PRIVATE` 且无此 header，link 时会拿不到。）
- **宏契约**：本 PR 用 `CPPLOG_INFO/WARN/ERROR(tag, fmt, ...)`。新 header 定义的正是这三个 printf 风格宏（转发到 `LogfSource(...)`），与 fallback shim 的 `FallbackLog(level, tag, format, ...)` 语义一致（都是 `vfprintf` 风格变参）✅。

### ✅ 其它核对
- **迁移完整性**：本文件内 `std::cerr`/`std::cout` 已全部清零（grep = 0），无残留。
- **签名/调用方**：三个 public 方法签名未变，唯一调用方 `hardware_simulator_plugin.cpp:1433/1437/1451` 不受影响。
- **一处顺带清理**：`DoControllerAction` 成功路径里那句语义错误的 `"GameController removed from slot"`（实际是 update 成功）被删掉了，合理。
- **格式化正确**：`0x%X` + `static_cast<unsigned int>(retval)` 替换了原来的 `std::hex <<`，等价且更适配变参 API。

### 💡 建议（均非阻塞，见 inline）
1. **fallback shim 前瞻性抽取** — 这 20 行 shim 目前 TU-local；后续迁移其它 4 个 Windows 源文件时会被逐份复制，建议提前抽到 `windows/cpp_log_shim.h`。
2. **fallback 原子性** — prefix 与 message 是两次 `stderr` 写，多线程下可能交错；`vsnprintf` 到栈 buffer 再单次 `fprintf` 即可。仅影响 standalone/example 构建，host 路径已由 ring buffer 串行化。

### ⚠️ 合并顺序提醒
本 PR 依赖 `cloudplayplus_log#2` 先合入并把 host 侧 submodule 指针更新到位；单独合本 PR 后若 host 的 `cpp_log_plugin` target 尚不存在，会走 `else()` 分支（native logging 静默禁用、退回 stderr）——不会 break 编译，但共享 logger 不生效。建议按 `cloudplayplus#407` 描述的顺序协同合并。

---

## 🧠 Claude review — **REQUEST_CHANGES**

## Code Review: `feat(logging): integrate shared native logger` (PR #36)

### Overview
The PR routes the Windows ViGEm game-controller diagnostics through CloudPlayPlus's shared native logger (`cpp_log`) instead of `std::cerr`/`std::cout`, with a `printf`-style stderr fallback (`FallbackLog` + `CPPLOG_*` macros) for standalone builds. The CMake guard links `cpp_log_plugin` and defines `CPP_LOG_AVAILABLE` when the host provides it. Direction is good and the fallback design is clean.

### What's good
- **Fallback semantics match the real logger.** The real API is `cpplog::Logf(Level, const char* tag, const char* fmt, ...)` (printf-style), and `CPPLOG_INFO(tag, ...)` etc. Your fallback macros mirror the exact `(tag, printf-format, args)` contract, so call sites are portable across both paths.
- **Correct format handling.** Switching `std::hex << retval` to `"0x%X"` with `static_cast<unsigned int>(...)` is right — `VIGEM_ERROR` is an enum, and `%X` needs an `unsigned int`.
- **Removed a copy-paste bug.** The stray `std::cout << "GameController removed from slot"` in `DoControllerAction` (which logged a removal on a successful *update*) is correctly deleted.

### Blocking concern — the native-logging path appears to never activate

Cross-checking against `CloudPlayPlus/cloudplayplus_log` (current HEAD), two wiring names in this PR don't match the actual package:

1. **Target name.** `cpp_log`'s `windows/CMakeLists.txt` defines its target as `add_library(cpp_log SHARED ...)` — it's a **dart:ffi plugin**, so Flutter's `generated_plugins.cmake` never creates a `cpp_log_plugin` target (that suffix only exists for method-channel plugins linked via `${plugin}_plugin`). So `if(TARGET cpp_log_plugin)` is always false → the `else` branch is always taken. The guard should be `if(TARGET cpp_log)`.

2. **Header path.** `#include <cpp_log/cpp_log.h>` doesn't exist. The macros/API live in `src/cpp_log_core.h` (README: "C++ — `src/cpp_log_core.h`"); the only header under `include/cpp_log/` is `cpp_log_apple.h`. There is no umbrella `cpp_log/cpp_log.h` anywhere in the org.

3. **Include propagation.** Even fixing (1), `cpp_log` does `target_include_directories(cpp_log PRIVATE "${SRC_DIR}")` — **PRIVATE**, so linking it wouldn't propagate the header path to this plugin. The logger package would need an INTERFACE/PUBLIC include dir (and ideally an umbrella header) for `#if defined(CPP_LOG_AVAILABLE)` to compile.

The insidious part: because the `else` branch is always selected, the **host Release build passes** (it just compiles the stderr fallback) even though the shared logger is never actually wired in — which is likely why validation looked green while the PR's stated goal (sharing the bounded async logger + rotating file sink) isn't met. As written, if `CPP_LOG_AVAILABLE` *were* ever defined, the include would fail to compile.

**Ask:** Confirm the real target name and public header. Most likely `if(TARGET cpp_log)` + an include path/umbrella header exported by the logger package (INTERFACE include dir). Please re-validate by confirming `CPP_LOG_AVAILABLE` is actually defined in the host build (e.g. the `message(STATUS ...)` does *not* fire), not just that the build succeeds.

### Minor / non-blocking
- `windows/gamecontroller_manager.h` still `#include <iostream>`, now unused since all stream I/O is gone. It also relies on `<iostream>` to transitively pull in `std::string` (used in `DoControllerAction`'s signature) — prefer an explicit `#include <string>` and drop `<iostream>`. Pre-existing, but this PR is the natural moment.
- `DoControllerAction`: `_XINPUT_GAMEPAD gamepad;` is left uninitialized and populated by unchecked `iss >>` extractions; a malformed `action` leaves fields indeterminate. Pre-existing and out of scope, but worth a follow-up.

### Verdict rationale
The diff itself is well-formed, but the CMake/header wiring names don't match the actual `cpp_log` package, so the feature silently degrades to the stderr fallback in the host build — the PR's core objective. That needs confirmation/fix before merge. If the maintainers verify a host-side shim provides `cpp_log_plugin` and `cpp_log/cpp_log.h`, this drops to COMMENT.

windows/gamecontroller_manager.cc
zhuhaichao518: 已在 65f11eb 处理：将 host logger include 与 standalone fallback 抽到 `windows/cpp_log_shim.h`，并由各 native TU 统一 include；同时把该头文件加入插件源文件列表，后续迁移其它 Windows 源文件时无需复制 shim。

windows/gamecontroller_manager.cc
zhuhaichao518: 已在 65f11eb 处理：fallback 先用 `vsnprintf` 格式化到栈 buffer，再通过单次 `fprintf` 写入完整日志行，避免 prefix/message 被并发写交错。standalone Windows Debug 和 host Release native target 均已重新编译通过。

zhuhaichao518:
Review feedback addressed in `65f11eb`:

- extracted the host/fallback logging adapter into shared `windows/cpp_log_shim.h`
- changed standalone fallback formatting to one complete `stderr` write
- replaced the obsolete `<iostream>` transitive dependency with explicit `<string>`
- confirmed the coordinated logger branch exports `cpp_log_plugin` and its public header; the host-generated Release project defines `CPP_LOG_AVAILABLE=1` and links that target

Validation: standalone Windows Debug build passed; host `cpp_log_plugin`, `hardware_simulator_plugin`, `flutter_webrtc_plugin`, and app Release targets compiled successfully.

zhuhaichao518:
@hxyshadow review

hxyshadow (approved):
**✅ APPROVE(通过)**

**✅ APPROVE**

## 复审（Case-A，前序 REQUEST_CHANGES → APPROVE）— #36 `feat(logging): integrate shared native logger`

**Prior reviewed:** `76bcf8e` ｜ **Now reviewed against:** `65f11eb`（作者推 fix commit + 重新 `@hxyshadow review`）。前轮我的 2 条 💡 与 claude 的 target-name 阻塞点**全部落地**，可以 APPROVE。

### 前轮建议核对（carry-forward，`76bcf8e..65f11eb`，4 文件）
1. ✅ **fallback shim 前瞻性抽取** → 已修。20 行 shim 从 `gamecontroller_manager.cc` 抽到独立头 `windows/cpp_log_shim.h`（namespace `hardware_simulator::logging`），并加进 `CMakeLists.txt` 的 `PLUGIN_SOURCES`。后续其它 Windows 源文件迁移可直接 `#include "cpp_log_shim.h"`。
2. ✅ **fallback 原子性** → 已修。新 shim 里 `FallbackLog` 先 `vsnprintf` 到栈 buffer `message[2048]`，再单次 `std::fprintf(stderr, "[%s] [%s] %s\n", ...)`——多线程下不再 prefix/message 两次写交错。
3. ✅ **（claude minor）`<iostream>` 未用** → 已修。`gamecontroller_manager.h` 删掉 `<iostream>`，显式加 `<string>`（`DoControllerAction` 签名需要），并顺手补了文件尾缺失的换行、修了 `DoControllerAction` 形参逗号空格。

### claude 前轮阻塞（target/header 名对不上 → native 静默降级）→ **已解**
claude REQUEST_CHANGES 的核心是 `if(TARGET cpp_log_plugin)` / `#include <cpp_log/cpp_log.h>` 对不上旧版 cpp_log。本轮对着 spdlog 重构后的 **cpp_log#2（`fc6c897`）**确认全部对上：
- ✅ cpp_log#2 `windows/CMakeLists.txt` 设 `PLUGIN_NAME="cpp_log_plugin"`（`add_library(cpp_log_plugin SHARED ...)`, `OUTPUT_NAME "cpp_log"`）——`if(TARGET cpp_log_plugin)` 名字对上。
- ✅ cpp_log#2 新增 `include/cpp_log/cpp_log.h`（printf 风格 `CPPLOG_INFO/WARN/ERROR(tag, ...)` → `LogfSource(...)`），且 `target_include_directories(... PUBLIC .../include)` 导出——`#include <cpp_log/cpp_log.h>` 可解析，宏契约与本 PR fallback shim 的 `CPPLOG_*` 语义一致。
- ✅ host `generated_plugins.cmake` 把 `cpp_log` 列进 `FLUTTER_PLUGIN_LIST`（idx2，先于本插件 idx8 配置）→ `cpp_log_plugin` target 在本插件配置时已存在 → `CPP_LOG_AVAILABLE=1` 真正定义，共享 logger 真接上（不再恒走 `else()` stderr 回落）。

### 迁移完整性 / 其它
- ✅ 本文件 `std::cerr`/`std::cout` 仍全部清零；三个 public 方法签名未变，唯一调用方 `hardware_simulator_plugin.cpp` 不受影响。

### 合并顺序提醒
本 PR 依赖 cpp_log#2 先合并并让 host 的 `cpp_log_plugin` target 就位；单独合本 PR 时若 host 尚无该 target，会走 `else()`（native logging 静默禁用、退回 stderr）——不 break 编译但共享 logger 不生效。建议按 #407 描述协同合并：runtime → 本 PR → host repoint。

---
🤖 Hermes reviewer · cron Case-A 复审

---

## Code Review: `feat(logging): integrate shared native logger` (PR #36)

### Overview
The PR routes the Windows ViGEm game-controller diagnostics through CloudPlayPlus's shared native logger (`cpp_log`), while keeping the plugin independently buildable:
- **CMake**: links `cpp_log_plugin` and defines `CPP_LOG_AVAILABLE=1` when the host provides that target; otherwise emits a status message and falls back.
- **`cpp_log_shim.h`** (new): includes the real `<cpp_log/cpp_log.h>` when available, else defines `CPPLOG_*` macros backed by a `vsnprintf`→`stderr` fallback.
- **`gamecontroller_manager.{cc,h}`**: replaces `std::cout`/`std::cerr` with `CPPLOG_*`, tidies includes (IWYU: drops `<iostream>`, adds `<string>`), fixes spacing and the missing trailing newline.

### Strengths
- **Clean fallback design.** The `#if defined(CPP_LOG_AVAILABLE)` split keeps standalone builds working, and the compile-time gate is scoped correctly to the plugin target.
- **No format-string injection.** All format strings are string literals; dynamic values (`retval`, slot ids) are passed as arguments, and error codes are cast to `unsigned int` for `%X`. Correct and safe.
- **IWYU cleanup is right.** `<string>` is now included directly (needed for `DoControllerAction(int, std::string&)`), and `<iostream>` removal is safe since the `.cc` no longer uses `cout`/`cerr`/`hex`/`endl`.
- **Incidental bug fix.** The old code logged `"GameController removed from slot"` on a successful *update* in `DoControllerAction` — a copy-paste artifact. This PR removes that misleading line and corrects the failure message to "update failed". Good.

### Issues / Risks
- **(Non-blocking, please confirm) External macro-contract assumption.** The `CPP_LOG_AVAILABLE` branch assumes `<cpp_log/cpp_log.h>` defines `CPPLOG_INFO/WARN/ERROR/...` as `(tag, format, ...)` printf-style macros. If the host's real API differs (different names, or no `tag` param, or `fmt`-style `{}` placeholders instead of `%X`), the host build breaks. The PR description says the host Release build was validated, so this is likely fine — just calling out that the shim and the host header are coupled by an unwritten contract. A one-line comment in `cpp_log_shim.h` documenting the expected signature would help future maintainers.
- **(Non-blocking) Transitive include propagation.** `target_link_libraries(... cpp_log_plugin)` only makes `<cpp_log/cpp_log.h>` resolvable if that target exports its include dir as `INTERFACE`/`PUBLIC`. Validated by the host build per the description; worth keeping in mind if the host target's include visibility ever changes.
- **(Minor) Silent 2048-byte truncation** in `FallbackLog`. Acceptable for diagnostics; `vsnprintf` bounds it safely.
- **(Scope note, not a defect)** Many sibling files (`virtual_display*.cc`, `SmartKeyboardBlocker.cpp`, `hardware_simulator_plugin.cpp`) still use raw `std::cout`/`std::cerr`. The PR explicitly scopes itself to "representative" diagnostics, so this is a deliberate incremental step — a follow-up to migrate the rest would be worthwhile.

### Tests
No unit tests added (consistent with the repo — this is native logging glue). Validation via standalone + host builds and `flutter analyze` is reasonable for this change.

### Verdict
Low-risk, well-structured incremental change. No correctness or security blockers found; the only open items are external-integration assumptions the author states were validated against the host build.

PR #40, opened 2026-07-29
fix(windows): move scroll scaling to Dart

zhuhaichao518:
## What changed

- Remove the Windows native `1.5` wheel magnitude multiplier.
- Keep the native adapter responsible only for axis direction, safe clamping, and Win32 injection.

## Why

The CloudPlayPlus Dart Host now applies source-platform-aware mouse-wheel normalization. Keeping the empirical multiplier in the native adapter would double-scale Windows input and prevents macOS mouse wheels from being normalized to a full Windows `WHEEL_DELTA` detent.

## Impact

This change must ship with the matching CloudPlayPlus client protocol update. It is intentionally not a standalone behavior change.

## Validation

- Parent CloudPlayPlus full Flutter test suite: 732 tests passed.
- Parent `flutter analyze --no-pub`: no issues.
- Windows native compilation still needs CI/Windows validation.

### DIFF
--- a/windows/hardware_simulator_plugin.cpp
+++ b/windows/hardware_simulator_plugin.cpp
@@ -51,9 +51,8 @@ constexpr ULONG_PTR kCursorReseedExtraInfo =
     static_cast<ULONG_PTR>(0x43505052);  // "CPPR"
 constexpr UINT_PTR kCursorReseedSubclassId =
     static_cast<UINT_PTR>(0x43505052);
-// Empirical tuning: maps Flutter page-scroll deltas to the desired Win32
-// wheel speed while preserving the logical protocol value on the wire.
-constexpr double kWindowsWheelSensitivity = 1.5;
+// Dart normalizes platform-specific scroll magnitudes. The native adapter only
+// converts the logical page direction and clamps it to a safe Win32 range.
 constexpr double kMaxWindowsWheelDistance =
     static_cast<double>(WHEEL_DELTA) * 100.0;
 
@@ -62,8 +61,7 @@ int logicalScrollToWindowsWheel(double logical_distance, bool invert) {
     return 0;
   }
   const double direction = invert ? -1.0 : 1.0;
-  const double scaled =
-      logical_distance * direction * kWindowsWheelSensitivity;
+  const double scaled = logical_distance * direction;
   const double clamped = (std::clamp)(
       scaled, -kMaxWindowsWheelDistance, kMaxWindowsWheelDistance);
   return static_cast<int>(std::lround(clamped));
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 * Define a cross-platform logical scroll contract and convert it at each
   native host boundary.
-* Reduce the empirical Windows scroll sensitivity from 2.0 to 1.5.
+* Move Windows scroll magnitude normalization to the Dart host layer; the
+  native adapter now applies direction and safety clamping only.
 
 ## 0.0.1
 


zhuhaichao518:
@hxyshadow 麻烦 review 一下这个 PR，谢谢！

windows/hardware_simulator_plugin.cpp
```diff
@@ -62,8 +61,7 @@ int logicalScrollToWindowsWheel(double logical_distance, bool invert) {
     return 0;
   }
   const double direction = invert ? -1.0 : 1.0;
-  const double scaled =
-      logical_distance * direction * kWindowsWheelSensitivity;
+  const double scaled = logical_distance * direction;
```
hxyshadow, line 64: 💡 **Suggestion:** 移除 `1.5` 系数后，`logicalScrollToWindowsWheel` 现在是纯粹的「方向 + clamp」变换，和 Linux 的 `logical_scroll_to_linux_wheel`、macOS 的 `nativeWheelDelta` 完全对齐 —— 这正是本 PR 想要的三端一致（Dart 侧统一 normalize）。

不过 Linux 侧已有针对该函数的单测（`linux/test/hardware_simulator_plugin_test.cc` 断言方向与 clamp 边界），Windows 侧目前没有对应覆盖（`windows/test/...` 只测了 `getPlatformVersion`）。本次刚好改动了这个函数的数值行为，建议补一个平行的单测（direction 正负、`0`、`±1e100` clamp 到 `±kMaxWindowsWheelDistance`），防止未来回归。非阻塞。

hxyshadow (commented):
**💬 COMMENT(评论)**

## Code Review: #40 — fix(windows): move scroll scaling to Dart

**Verdict: Approved ✅** (0 issues, 2 suggestions)
**Author:** @zhuhaichao518 | **Files:** 1 (+3 -5) | Reviewed against: `e72e552`

移除 Windows 原生的 `1.5` 滚轮系数，让 `logicalScrollToWindowsWheel` 只做「方向 + 安全 clamp」，把滚动幅度的 normalize 全部交给 Dart Host。改动干净、局部、方向正确。

### 跨文件 / 跨平台校验
- `kWindowsWheelSensitivity`：全仓 **0 处残留引用**，删除干净。
- 函数签名不变，两处调用点（`performMouseScroll` @ 1387-1388，同文件）不受影响。
- **三端对称性（关键）**：macOS `nativeWheelDelta`（无系数，`(-logicalDelta).rounded()` + clamp）与 Linux `logical_scroll_to_linux_wheel`（无系数，`delta*direction` + clamp）本来就是 1:1；本 PR 把 Windows 也拉齐到 1:1 —— 正是 PR 描述里「Dart 统一 normalize、native 只转方向」的目标。symmetry 达成。
- 符合 PR 说明：这是协同改动，需与 CloudPlayPlus 客户端协议更新一起上线，本身不是独立行为变更。

### 💡 Suggestions
- **CHANGELOG.md（Unreleased）** — 仍写着「Reduce the empirical Windows scroll sensitivity from 2.0 to 1.5」，但本 PR 是把 `1.5` 系数**整个删除**，这行现在与代码矛盾。建议补一条说明「移除 Windows 经验系数、幅度归一交由 Dart」并修掉这行旧描述。
- **windows/hardware_simulator_plugin.cpp:64** — Windows 侧缺少对 `logicalScrollToWindowsWheel` 的单测（Linux 已有平行单测覆盖方向与 clamp）。本次改了它的数值行为，建议补齐平行测试，防回归。（inline 已标）

### ✅ Looks Good
- 系数删除彻底、无残留引用；三端滚动语义收敛到单一真相源（Dart normalize）。
- `std::isfinite` 保护与 clamp 边界保留完好，方向注释清晰准确。

（注：PR 目前为 draft，作者已 @ 请求 review。以上两条均非阻塞。）

---

## Overview

Removes the `kWindowsWheelSensitivity = 1.5` empirical multiplier from `windows/hardware_simulator_plugin.cpp`, so `logicalScrollToWindowsWheel()` now only applies direction inversion + clamping. Magnitude normalization moves up to the Dart host. Net: −5/+3 in one file.

## Correctness

- The math change is straightforward and correct: `scaled = logical_distance * direction`. Finite-check, clamp, and `lround` are unchanged, so no new overflow/UB path — `kMaxWindowsWheelDistance` (WHEEL_DELTA × 100) still bounds the value well within `int`.
- Removing the constant is complete; no other reference to `kWindowsWheelSensitivity` remains in the tree.
- The comment rewrite accurately describes the new contract and is placed sensibly, though it now sits directly above `kMaxWindowsWheelDistance` where it partly reads as that constant's doc. Minor: consider moving it to a doc comment on `logicalScrollToWindowsWheel` instead.

## Issues to address

1. **Stale CHANGELOG (should fix).** `CHANGELOG.md` under `## Unreleased` still says:
   > * Reduce the empirical Windows scroll sensitivity from 2.0 to 1.5.

   That line is now false — the multiplier is gone entirely. Since it's still unreleased, replace it with something like "Move Windows scroll magnitude normalization to the Dart layer; native applies direction and clamping only." Leaving it would ship a changelog that contradicts the code.

2. **Cross-repo coupling is not enforceable here.** The PR body correctly notes this must ship with the matching CloudPlayPlus client update, but `hardware_simulator` is a standalone pub package (`version: 0.0.1`). Any other consumer that calls the method channel directly gets a silent 1.5× scroll regression with no compile error and no version signal. Recommend at minimum documenting the logical-scroll contract (units: what does `logical_distance == 1.0` mean?) in the public Dart API docs, so downstream callers know they own normalization. A version bump when this is released would also help.

3. **macOS asymmetry.** `macos/Classes/HardwareSimulatorPlugin.swift` posts with `units: .pixel` and no scaling, while Windows now emits raw logical → `WHEEL_DELTA` units. Those two are only equivalent if Dart emits per-platform-target values. Worth confirming the Dart side keys normalization on the *target* platform, not just the source — the PR body mentions "source-platform-aware," which is only half the mapping.

## Tests

- `windows/test/hardware_simulator_plugin_test.cpp` has no wheel/scroll coverage, so nothing breaks — but nothing guards this either. `logicalScrollToWindowsWheel` is a pure function and an easy unit-test target (direction, invert, clamp bounds, non-finite → 0). Adding 4–5 assertions would make the next tuning change safe. Not blocking, but this is the second time this constant has moved.
- Validation section cites the parent repo's suite; nothing in *this* repo exercises the change. Windows CI compilation check is still pending as the author notes.

## Performance / Security

No concerns. One fewer multiply on an input-event path; no new syscalls, allocations, or unvalidated input. Clamping still protects the Win32 injection.

## Style

Consistent with the surrounding file — `constexpr`, `k`-prefixed naming, 2-space indent, comment tone all match.

Overall a clean, minimal, well-motivated change. The only thing I'd want fixed before merge is the CHANGELOG line; the rest are recommendations.

zhuhaichao518:
已在 16c29a0 更新 `CHANGELOG.md`，删除与当前实现矛盾的 Windows 1.5 倍描述，改为明确 native 仅负责方向与安全钳位、幅度归一由 Dart Host 负责。本 PR 保持最小范围，Windows 单测和公共 API 单位契约建议后续单独补充。@hxyshadow 麻烦再看一下，谢谢！

hxyshadow (commented):
**💬 COMMENT(评论)**

## Overview

Removes the empirical `kWindowsWheelSensitivity = 1.5` multiplier from the Windows native wheel path, leaving the adapter responsible only for direction inversion and clamping. Magnitude normalization moves up to the Dart host. CHANGELOG updated accordingly. 5 additions / 6 deletions across 2 files.

## Correctness

- The remaining math is sound: `logical_distance * direction`, clamped to ±`WHEEL_DELTA * 100`, then `std::lround`. Removing a constant factor can't introduce overflow or sign issues, and the early-return-on-zero guard is untouched.
- Semantics of `logical_distance` are now "already in Win32 wheel units" rather than "logical pages." That's a meaningful contract change encoded only in a comment. It's the correct direction (single source of truth in Dart), but it's a **cross-repo coupling**: this plugin version is only correct paired with the updated CloudPlayPlus client. The PR body acknowledges this.

## Suggestions

- **Version the break.** Since the wire meaning of the value changed while the API shape didn't, an older Dart caller against a newer plugin silently scrolls at 2/3 speed with no error. A `pubspec.yaml` version bump (and ideally a `## 0.0.2` heading in CHANGELOG — the new entries currently sit above `## 0.0.1` under an unlabeled section) would let the client pin the pairing.
- **Document the unit in the signature.** `logicalScrollToWindowsWheel(double logical_distance, ...)` — the parameter name still says "logical," but post-change it's expected in `WHEEL_DELTA` units. Renaming to `wheel_delta_distance` or stating the unit in the comment above the function (not just at the constant) would prevent the next reader from re-adding a scale factor.
- Minor: with no scaling left, the function is nearly `lround(clamp(±x))`. Fine to keep as-is for the clamping/inversion, but the comment explaining *why* it's now a near-identity is doing important work — good that you added it.

## Tests / performance / security

- No native tests added or changed; the validation section cites the parent repo's Dart suite and notes Windows compilation is unverified. A unit test on `logicalScrollToWindowsWheel` (direction, clamp boundary, zero, rounding) would be cheap insurance for a function whose contract just shifted, but the codebase appears to have no existing native test harness, so this is a suggestion rather than a gap in convention.
- No performance or security implications — arithmetic only, clamping retained so no unbounded value reaches `mouse_event`/`SendInput`.

Overall this is a clean, well-motivated simplification with an accurate changelog. My only real concern is the unversioned cross-repo contract change, which is a process/release matter rather than a defect in the code — and the PR is already merged.

PR #4, opened 2025-07-15
Add Display and config resolution.

HXYerror:
- [x] 1. Keep displays live
- [x] 2. Config display resolution
- [x] 3. Customize resolution
- [x] 4. Configure more properties
image

 ---------- Next pr
- [ ] 1. 优化更新display的设计
- [ ] 2. 非admin提权设置注册表
- [ ] 3. 多显示器 >2 拓扑结构设置，复杂

PR #41, opened 2026-07-29
feat(macos): expose native trackpad scroll events

zhuhaichao518:
## What changed

- add a cross-platform `TrackpadScrollEvent` callback contract with Flutter-view position, canonical page-direction deltas, gesture phase, and momentum state
- start native capture for the first listener and stop it after the last listener is removed
- implement macOS capture from precise `NSEvent.scrollWheel` events without consuming the AppKit event
- add a unified precise-trackpad injection API that preserves fractional pixels where the platform supports it
- on macOS, rebuild continuous CoreGraphics events with 16.16 fixed-point deltas plus public scroll/momentum phase fields
- on Windows 10, replay unified trackpad input as ordinary high-resolution `WHEEL/HWHEEL` events and retain per-axis fractional remainders at the integer API boundary
- keep all magnitude scaling in Viewer Dart; Windows native applies no `×1.2`, `×1.5`, `×2`, mouse sensitivity multiplier, or `lround` on trackpad input
- keep Windows native limited to axis direction, safe clamping, fractional remainder carry and Win32 integer injection
- keep ordinary mouse-wheel magnitude unchanged and unsupported capture platforms as safe no-ops

## Why

Flutter macOS `PointerPanZoom` does not preserve the complete native momentum sequence. Reusing integer mouse-wheel injection also rounds native values such as `0.1` to zero on macOS. Native capture and precise macOS injection keep fractional pixels, direction conversion, and lifecycle metadata in the platform plugin. Windows 10 has no public precision-touchpad injector, so it replays Viewer-normalized values through ordinary wheel input without changing their cumulative magnitude.

## Validation

- plugin `flutter test --no-pub` (8 tests)
- Windows accumulator tests cover 1:1 integral distance and fractional remainder carry
- parent app 63 focused Viewer/normalizer tests
- parent app `flutter analyze --no-pub`
- parent app signed macOS release build with `CPP_LOG_AVAILABLE`
- verified against parent `cloudplayplus#442` commit `96dd510`

### DIFF
--- a/lib/hardware_simulator_method_channel.dart
+++ b/lib/hardware_simulator_method_channel.dart
@@ -45,7 +45,9 @@ class MethodChannelHardwareSimulator extends HardwareSimulatorPlatform {
             phase: _trackpadScrollPhase(arguments['phase']),
             isMomentum: arguments['isMomentum'] == true,
           );
-          for (var callback in trackpadScrollCallbacks) {
+          for (final callback in List<TrackpadScrollCallback>.of(
+            trackpadScrollCallbacks,
+          )) {
             callback(event);
           }
         }
@@ -263,6 +265,7 @@ class MethodChannelHardwareSimulator extends HardwareSimulatorPlatform {
   @override
   void addTrackpadScroll(TrackpadScrollCallback callback) {
     if (!isinitialized) init();
+    if (trackpadScrollCallbacks.contains(callback)) return;
     final shouldStartCapture = trackpadScrollCallbacks.isEmpty;
     trackpadScrollCallbacks.add(callback);
     if (shouldStartCapture) {
--- a/lib/hardware_simulator_method_channel.dart
+++ b/lib/hardware_simulator_method_channel.dart
@@ -34,6 +34,21 @@ class MethodChannelHardwareSimulator extends HardwareSimulatorPlatform {
         for (var callback in cursorWheelCallbacks) {
           callback(call.arguments['dx'], call.arguments['dy']);
         }
+      } else if (call.method == "onTrackpadScroll") {
+        final arguments = call.arguments;
+        if (arguments is Map) {
+          final event = TrackpadScrollEvent(
+            x: _asDouble(arguments['x']),
+            y: _asDouble(arguments['y']),
+            deltaX: _asDouble(arguments['dx']),
+            deltaY: _asDouble(arguments['dy']),
+            phase: _trackpadScrollPhase(arguments['phase']),
+            isMomentum: arguments['isMomentum'] == true,
+          );
+          for (var callback in trackpadScrollCallbacks) {
+            callback(event);
+          }
+        }
       } else if (call.method == "onCursorImageMessage") {
         int callbackID = call.arguments['callbackID'];
         if (cursorImageCallbacks.containsKey(callbackID)) {
@@ -243,6 +258,46 @@ class MethodChannelHardwareSimulator extends HardwareSimulatorPlatform {
     cursorWheelCallbacks.remove(callback);
   }
 
+  final List<TrackpadScrollCallback> trackpadScrollCallbacks = [];
+
+  @override
+  void addTrackpadScroll(TrackpadScrollCallback callback) {
+    if (!isinitialized) init();
+    final shouldStartCapture = trackpadScrollCallbacks.isEmpty;
+    trackpadScrollCallbacks.add(callback);
+    if (shouldStartCapture) {
+      unawaited(startTrackpadScrollCapture());
+    }
+  }
+
+  @override
+  void removeTrackpadScroll(TrackpadScrollCallback callback) {
+    final removed = trackpadScrollCallbacks.remove(callback);
+    if (removed && trackpadScrollCallbacks.isEmpty) {
+      unawaited(stopTrackpadScrollCapture());
+    }
+  }
+
+  @override
+  Future<bool> startTrackpadScrollCapture() async {
+    try {
+      final supported = await methodChannel
+          .invokeMethod<Object?>('startTrackpadScrollCapture');
+      return supported == true;
+    } on MissingPluginException {
+      return false;
+    }
+  }
+
+  @override
+  Future<void> stopTrackpadScrollCapture() async {
+    try {
+      await methodChannel.invokeMethod<void>('stopTrackpadScrollCapture');
+    } on MissingPluginException {
+      return;
+    }
+  }
+
   final Map<int, CursorImageUpdatedCallback> cursorImageCallbacks = {};
   final Map<int, CursorPositionUpdatedCallback> cursorPositionCallbacks = {};
   final Map<int, DisplayCountChangedCallback> displayCountCallbacks = {};
@@ -373,6 +428,22 @@ class MethodChannelHardwareSimulator extends HardwareSimulatorPlatform {
       'dx': dx,
       'dy': dy,
     });
+  }
+
+  static double _asDouble(Object? value) {
+    return value is num ? value.toDouble() : 0;
+  }
+
+  static TrackpadScrollPhase _trackpadScrollPhase(Object? value) {
+    return switch (value) {
+      'mayBegin' => TrackpadScrollPhase.mayBegin,
+      'began' => TrackpadScrollPhase.began,
+      'changed' => TrackpadScrollPhase.changed,
+      'stationary' => TrackpadScrollPhase.stationary,
+      'ended' => TrackpadScrollPhase.ended,
+      'cancelled' => TrackpadScrollPhase.cancelled,
+      _ => TrackpadScrollPhase.none,
+    };
   }
 
   @override
--- a/lib/hardware_simulator_method_channel.dart
+++ b/lib/hardware_simulator_method_channel.dart
@@ -440,12 +440,16 @@ class MethodChannelHardwareSimulator extends HardwareSimulatorPlatform {
     TrackpadScrollPhase phase = TrackpadScrollPhase.none,
     bool isMomentum = false,
   }) async {
-    await methodChannel.invokeMethod('trackpadScroll', {
-      'dx': dx,
-      'dy': dy,
-      'phase': phase.name,
-      'isMomentum': isMomentum,
-    });
+    try {
+      await methodChannel.invokeMethod('trackpadScroll', {
+        'dx': dx,
+        'dy': dy,
+        'phase': phase.name,
+        'isMomentum': isMomentum,
+      });
+    } on MissingPluginException {
+      await performMouseScroll(dx, dy);
+    }
   }
 
   static double _asDouble(Object? value) {
--- a/lib/hardware_simulator_method_channel.dart
+++ b/lib/hardware_simulator_method_channel.dart
@@ -433,6 +433,21 @@ class MethodChannelHardwareSimulator extends HardwareSimulatorPlatform {
     });
   }
 
+  @override
+  Future<void> performTrackpadScroll(
+    double dx,
+    double dy, {
+    TrackpadScrollPhase phase = TrackpadScrollPhase.none,
+    bool isMomentum = false,
+  }) async {
+    await methodChannel.invokeMethod('trackpadScroll', {
+      'dx': dx,
+      'dy': dy,
+      'phase': phase.name,
+      'isMomentum': isMomentum,
+    });
+  }
+
   static double _asDouble(Object? value) {
     return value is num ? value.toDouble() : 0;
   }
--- a/macos/Classes/HardwareSimulatorPlugin.swift
+++ b/macos/Classes/HardwareSimulatorPlugin.swift
@@ -15,6 +15,8 @@ class CursorConstants {
 
 public class HardwareSimulatorPlugin: NSObject, FlutterPlugin {
   private var methodChannel: FlutterMethodChannel?
+  private weak var flutterView: NSView?
+  private var trackpadScrollMonitor: Any?
   private var defaultCursorHasher: CursorHasher?
   private var currentScreenId: Int = 0
   private var lastMouseClickButtonId: Int?
@@ -57,6 +59,7 @@ public class HardwareSimulatorPlugin: NSObject, FlutterPlugin {
     let channel = FlutterMethodChannel(name: "hardware_simulator", binaryMessenger: registrar.messenger)
     let instance = HardwareSimulatorPlugin()
     instance.methodChannel = channel
+    instance.flutterView = registrar.view
     instance.defaultCursorHasher = CursorHasher()
 #if CLOUDPLAYPLUS_DMG_DISTRIBUTION
     instance.registerMacTerminationObserver()
@@ -65,12 +68,65 @@ public class HardwareSimulatorPlugin: NSObject, FlutterPlugin {
   }
 
   deinit {
+    stopTrackpadScrollCapture()
 #if CLOUDPLAYPLUS_DMG_DISTRIBUTION
     if let macTerminationObserver {
       NotificationCenter.default.removeObserver(macTerminationObserver)
     }
     restoreMacDisplaysBeforeExit()
 #endif
+  }
+
+  private func startTrackpadScrollCapture() {
+    guard trackpadScrollMonitor == nil else { return }
+    trackpadScrollMonitor = NSEvent.addLocalMonitorForEvents(
+      matching: .scrollWheel
+    ) { [weak self] event in
+      self?.handleTrackpadScroll(event)
+      return event
+    }
+  }
+
+  private func stopTrackpadScrollCapture() {
+    guard let trackpadScrollMonitor else { return }
+    NSEvent.removeMonitor(trackpadScrollMonitor)
+    self.trackpadScrollMonitor = nil
+  }
+
+  private func handleTrackpadScroll(_ event: NSEvent) {
+    guard event.hasPreciseScrollingDeltas else { return }
+    guard let flutterView, event.window === flutterView.window else { return }
+
+    let isMomentum = event.momentumPhase.rawValue != 0
+    let eventPhase = isMomentum ? event.momentumPhase : event.phase
+    guard eventPhase.rawValue != 0 else { return }
+
+    let point = flutterView.convert(event.locationInWindow, from: nil)
+    let flutterY = flutterView.isFlipped
+      ? point.y
+      : flutterView.bounds.height - point.y
+    methodChannel?.invokeMethod(
+      "onTrackpadScroll",
+      arguments: [
+        "x": point.x,
+        "y": flutterY,
+        // AppKit reports gesture direction; the plugin API uses page direction.
+        "dx": -event.scrollingDeltaX,
+        "dy": -event.scrollingDeltaY,
+        "phase": trackpadScrollPhaseName(eventPhase),
+        "isMomentum": isMomentum,
+      ]
+    )
+  }
+
+  private func trackpadScrollPhaseName(_ phase: NSEvent.Phase) -> String {
+    if phase.contains(.cancelled) { return "cancelled" }
+    if phase.contains(.ended) { return "ended" }
+    if phase.contains(.began) { return "began" }
+    if phase.contains(.mayBegin) { return "mayBegin" }
+    if phase.contains(.stationary) { return "stationary" }
+    if phase.contains(.changed) { return "changed" }
+    return "none"
   }
 
 #if CLOUDPLAYPLUS_DMG_DISTRIBUTION
@@ -2305,6 +2361,12 @@ public class HardwareSimulatorPlugin: NSObject, FlutterPlugin {
     switch call.method {
     case "getPlatformVersion":
       result("macOS " + ProcessInfo.processInfo.operatingSystemVersionString)
+    case "startTrackpadScrollCapture":
+      startTrackpadScrollCapture()
+      result(true)
+    case "stopTrackpadScrollCapture":
+      stopTrackpadScrollCapture()
+      result(nil)
     case "checkMacOSPermissions":
       // Reports the *current process's* live TCC status. Does NOT prompt.
       // - screenCapture: needed to grab the screen (ScreenCaptureKit / CGDisplayStream)
--- a/macos/Classes/HardwareSimulatorPlugin.swift
+++ b/macos/Classes/HardwareSimulatorPlugin.swift
@@ -1935,6 +1935,114 @@ public class HardwareSimulatorPlugin: NSObject, FlutterPlugin {
       }
   }
 
+  private func nativeTrackpadPointDelta(_ logicalDelta: Double) -> Int32? {
+      guard logicalDelta.isFinite else { return nil }
+      if logicalDelta == 0 { return 0 }
+
+      // A local macOS event measured as scrollingDelta=0.1 carries a
+      // wheel/point axis of one and a 16.16 fixed-point delta of 0.1.
+      // Rebuild both representations instead of rounding the precise value.
+      var rounded = (logicalDelta * 10).rounded()
+      if rounded == 0 {
+          rounded = logicalDelta.isLess(than: 0) ? -1 : 1
+      }
+      let clamped = min(
+          max(rounded, Double(Int32.min)),
+          Double(Int32.max)
+      )
+      return Int32(clamped)
+  }
+
+  private func nativeTrackpadScrollPhase(_ phase: String) -> Int64 {
+      switch phase {
+      case "mayBegin":
+          return Int64(CGScrollPhase.mayBegin.rawValue)
+      case "began":
+          return Int64(CGScrollPhase.began.rawValue)
+      case "changed", "stationary":
+          return Int64(CGScrollPhase.changed.rawValue)
+      case "ended":
+          return Int64(CGScrollPhase.ended.rawValue)
+      case "cancelled":
+          return Int64(CGScrollPhase.cancelled.rawValue)
+      default:
+          return 0
+      }
+  }
+
+  private func nativeTrackpadMomentumPhase(_ phase: String) -> Int64 {
+      switch phase {
+      case "mayBegin", "began":
+          return Int64(CGMomentumScrollPhase.begin.rawValue)
+      case "changed", "stationary":
+          return Int64(CGMomentumScrollPhase.continuous.rawValue)
+      case "ended", "cancelled":
+          return Int64(CGMomentumScrollPhase.end.rawValue)
+      default:
+          return Int64(CGMomentumScrollPhase.none.rawValue)
+      }
+  }
+
+  func performTrackpadScroll(
+      dx: Double,
+      dy: Double,
+      phase: String,
+      isMomentum: Bool
+  ) {
+      guard
+          let horizontalPoint = nativeTrackpadPointDelta(dx),
+          let verticalPoint = nativeTrackpadPointDelta(dy)
+      else {
+          return
+      }
+
+      let eventSource = CGEventSource(stateID: .hidSystemState)
+      let wheelCount: UInt32 = horizontalPoint != 0 ? 2 : 1
+      guard let scrollEvent = CGEvent(
+          scrollWheelEvent2Source: eventSource,
+          units: .pixel,
+          wheelCount: wheelCount,
+          wheel1: verticalPoint,
+          wheel2: horizontalPoint,
+          wheel3: 0
+      ) else {
+          return
+      }
+
+      // RD trackpad deltas use logical page direction. A native macOS precise
+      // event carries the opposite gesture-direction value in its 16.16
+      // fixed-point fields, while wheel/point axes retain page direction.
+      scrollEvent.setDoubleValueField(
+          .scrollWheelEventFixedPtDeltaAxis1,
+          value: -dy
+      )
+      scrollEvent.setDoubleValueField(
+          .scrollWheelEventFixedPtDeltaAxis2,
+          value: -dx
+      )
+      scrollEvent.setIntegerValueField(
+          .scrollWheelEventPointDeltaAxis1,
+          value: Int64(verticalPoint)
+      )
+      scrollEvent.setIntegerValueField(
+          .scrollWheelEventPointDeltaAxis2,
+          value: Int64(horizontalPoint)
+      )
+      scrollEvent.setIntegerValueField(
+          .scrollWheelEventIsContinuous,
+          value: 1
+      )
+      scrollEvent.setIntegerValueField(
+          .scrollWheelEventScrollPhase,
+          value: isMomentum ? 0 : nativeTrackpadScrollPhase(phase)
+      )
+      scrollEvent.setIntegerValueField(
+          .scrollWheelEventMomentumPhase,
+          value: isMomentum ? nativeTrackpadMomentumPhase(phase) : 0
+      )
+      scrollEvent.post(tap: .cghidEventTap)
+  }
+
   func performMouseMoveToWindowPosition(percentx: Double, percenty: Double) {
       // Get the current Flutter app's main window
       guard let mainWindow = NSApplication.shared.mainWindow ?? NSApplication.shared.windows.first else {
@@ -2643,6 +2751,22 @@ public class HardwareSimulatorPlugin: NSObject, FlutterPlugin {
         result(nil) // 表示成功执行，不返回值
       } else {
         result(FlutterError(code: "BAD_ARGS", message: "Missing or incorrect arguments for Mouse Scroll", details: nil))
+      }
+    case "trackpadScroll":
+      if let args = call.arguments as? [String: Any],
+        let dx = args["dx"] as? Double,
+        let dy = args["dy"] as? Double,
+        let phase = args["phase"] as? String,
+        let isMomentum = args["isMomentum"] as? Bool {
+        performTrackpadScroll(
+          dx: dx,
+          dy: dy,
+          phase: phase,
+          isMomentum: isMomentum
+        )
+        result(nil)
+      } else {
+        result(FlutterError(code: "BAD_ARGS", message: "Missing or incorrect arguments for Trackpad Scroll", details: nil))
       }
     case "mouseMoveToWindowPosition":
       if let args = call.arguments as? [String: Any],
--- a/macos/Classes/HardwareSimulatorPlugin.swift
+++ b/macos/Classes/HardwareSimulatorPlugin.swift
@@ -1935,16 +1935,16 @@ public class HardwareSimulatorPlugin: NSObject, FlutterPlugin {
       }
   }
 
-  private func nativeTrackpadPointDelta(_ logicalDelta: Double) -> Int32? {
+  static func nativeTrackpadPointDelta(_ logicalDelta: Double) -> Int32? {
       guard logicalDelta.isFinite else { return nil }
       if logicalDelta == 0 { return 0 }
 
-      // A local macOS event measured as scrollingDelta=0.1 carries a
-      // wheel/point axis of one and a 16.16 fixed-point delta of 0.1.
-      // Rebuild both representations instead of rounding the precise value.
-      var rounded = (logicalDelta * 10).rounded()
+      // CoreGraphics wheel/point axes use the opposite sign from our logical
+      // page direction. Preserve their 1:1 magnitude for whole pixels; only
+      // clamp a non-zero subpixel delta to the smallest representable integer.
+      var rounded = (-logicalDelta).rounded()
       if rounded == 0 {
-          rounded = logicalDelta.isLess(than: 0) ? -1 : 1
+          rounded = logicalDelta.isLess(than: 0) ? 1 : -1
       }
       let clamped = min(
           max(rounded, Double(Int32.min)),
@@ -1990,8 +1990,8 @@ public class HardwareSimulatorPlugin: NSObject, FlutterPlugin {
       isMomentum: Bool
   ) {
       guard
-          let horizontalPoint = nativeTrackpadPointDelta(dx),
-          let verticalPoint = nativeTrackpadPointDelta(dy)
+          let horizontalPoint = Self.nativeTrackpadPointDelta(dx),
+          let verticalPoint = Self.nativeTrackpadPointDelta(dy)
       else {
           return
       }
@@ -2009,9 +2009,9 @@ public class HardwareSimulatorPlugin: NSObject, FlutterPlugin {
           return
       }
 
-      // RD trackpad deltas use logical page direction. A native macOS precise
-      // event carries the opposite gesture-direction value in its 16.16
-      // fixed-point fields, while wheel/point axes retain page direction.
+      // RD trackpad deltas use logical page direction. CoreGraphics fields use
+      // the opposite gesture/device direction. Fixed-point preserves the exact
+      // fraction; point/wheel fields provide the closest non-zero integer.
       scrollEvent.setDoubleValueField(
           .scrollWheelEventFixedPtDeltaAxis1,
           value: -dy
--- a/macos/Classes/HardwareSimulatorPlugin.swift
+++ b/macos/Classes/HardwareSimulatorPlugin.swift
@@ -15,6 +15,7 @@ class CursorConstants {
 
 public class HardwareSimulatorPlugin: NSObject, FlutterPlugin {
   private var methodChannel: FlutterMethodChannel?
+  private weak var registrar: FlutterPluginRegistrar?
   private weak var flutterView: NSView?
   private var trackpadScrollMonitor: Any?
   private var defaultCursorHasher: CursorHasher?
@@ -59,7 +60,7 @@ public class HardwareSimulatorPlugin: NSObject, FlutterPlugin {
     let channel = FlutterMethodChannel(name: "hardware_simulator", binaryMessenger: registrar.messenger)
     let instance = HardwareSimulatorPlugin()
     instance.methodChannel = channel
-    instance.flutterView = registrar.view
+    instance.registrar = registrar
     instance.defaultCursorHasher = CursorHasher()
 #if CLOUDPLAYPLUS_DMG_DISTRIBUTION
     instance.registerMacTerminationObserver()
@@ -77,14 +78,19 @@ public class HardwareSimulatorPlugin: NSObject, FlutterPlugin {
 #endif
   }
 
-  private func startTrackpadScrollCapture() {
-    guard trackpadScrollMonitor == nil else { return }
+  private func startTrackpadScrollCapture() -> Bool {
+    if trackpadScrollMonitor != nil {
+      return flutterView != nil
+    }
+    flutterView = registrar?.view ?? registrar?.viewController?.view
+    guard flutterView != nil else { return false }
     trackpadScrollMonitor = NSEvent.addLocalMonitorForEvents(
       matching: .scrollWheel
     ) { [weak self] event in
       self?.handleTrackpadScroll(event)
       return event
     }
+    return true
   }
 
   private func stopTrackpadScrollCapture() {
@@ -95,6 +101,9 @@ public class HardwareSimulatorPlugin: NSObject, FlutterPlugin {
 
   private func handleTrackpadScroll(_ event: NSEvent) {
     guard event.hasPreciseScrollingDeltas else { return }
+    if flutterView?.window !== event.window {
+      flutterView = registrar?.view ?? registrar?.viewController?.view
+    }
     guard let flutterView, event.window === flutterView.window else { return }
 
     let isMomentum = event.momentumPhase.rawValue != 0
@@ -2362,8 +2371,7 @@ public class HardwareSimulatorPlugin: NSObject, FlutterPlugin {
     case "getPlatformVersion":
       result("macOS " + ProcessInfo.processInfo.operatingSystemVersionString)
     case "startTrackpadScrollCapture":
-      startTrackpadScrollCapture()
-      result(true)
+      result(startTrackpadScrollCapture())
     case "stopTrackpadScrollCapture":
       stopTrackpadScrollCapture()
       result(nil)
--- a/macos/Classes/HardwareSimulatorPlugin.swift
+++ b/macos/Classes/HardwareSimulatorPlugin.swift
@@ -94,9 +94,11 @@ public class HardwareSimulatorPlugin: NSObject, FlutterPlugin {
   }
 
   private func stopTrackpadScrollCapture() {
-    guard let trackpadScrollMonitor else { return }
-    NSEvent.removeMonitor(trackpadScrollMonitor)
+    if let trackpadScrollMonitor {
+      NSEvent.removeMonitor(trackpadScrollMonitor)
+    }
     self.trackpadScrollMonitor = nil
+    flutterView = nil
   }
 
   private func handleTrackpadScroll(_ event: NSEvent) {
--- a/windows/hardware_simulator_plugin.cpp
+++ b/windows/hardware_simulator_plugin.cpp
@@ -56,12 +56,6 @@ constexpr UINT_PTR kCursorReseedSubclassId =
 // converts the logical page direction and clamps it to a safe Win32 range.
 constexpr double kMaxWindowsWheelDistance =
     static_cast<double>(WHEEL_DELTA) * 100.0;
-// Chromium's legacy Windows wheel path maps one 120-unit detent to 100 CSS
-// pixels with the default three-line setting. Convert native macOS trackpad
-// pixels into that Win32 unit space so cumulative page movement stays 1:1.
-constexpr double kWindowsTrackpadWheelUnitsPerPixel =
-    static_cast<double>(WHEEL_DELTA) / 100.0;
-
 int logicalScrollToWindowsWheel(double logical_distance, bool invert) {
   if (!std::isfinite(logical_distance)) {
     return 0;
@@ -1405,11 +1399,11 @@ void performTrackpadScroll(double dx, double dy) {
     // Windows 10 has no public precision-touchpad injection API. Replay each
     // native trackpad frame as an ordinary high-resolution wheel event. The
     // legacy API is integral, so each axis carries its fractional wheel-unit
-    // remainder forward to preserve cumulative page distance.
+    // remainder forward. Cross-platform magnitude scaling belongs to Dart.
     thread_local TrackpadScrollAccumulator horizontal_accumulator(
-        kWindowsTrackpadWheelUnitsPerPixel, kMaxWindowsWheelDistance);
+        kMaxWindowsWheelDistance);
     thread_local TrackpadScrollAccumulator vertical_accumulator(
-        kWindowsTrackpadWheelUnitsPerPixel, kMaxWindowsWheelDistance);
+        kMaxWindowsWheelDistance);
     const int horizontal_distance =
         horizontal_accumulator.Convert(dx, false);
     const int vertical_distance =
--- a/windows/hardware_simulator_plugin.cpp
+++ b/windows/hardware_simulator_plugin.cpp
@@ -67,6 +67,22 @@ int logicalScrollToWindowsWheel(double logical_distance, bool invert) {
   return static_cast<int>(std::lround(clamped));
 }
 
+int logicalTrackpadScrollToWindowsWheel(
+    double logical_distance,
+    bool invert) {
+  if (!std::isfinite(logical_distance)) {
+    return 0;
+  }
+  const double direction = invert ? -1.0 : 1.0;
+  const double converted = logical_distance * direction;
+  const double clamped = (std::clamp)(
+      converted, -kMaxWindowsWheelDistance, kMaxWindowsWheelDistance);
+  // Win32 MOUSEINPUT::mouseData is integral. The trackpad path deliberately
+  // sends the canonical 1x delta directly at that API boundary: no mouse-wheel
+  // sensitivity multiplier, fractional accumulator, or nearest-value rounding.
+  return static_cast<int>(clamped);
+}
+
 LRESULT CALLBACK CursorReseedSubclassProc(
     HWND window,
     UINT message,
@@ -1395,6 +1411,23 @@ void performMouseScroll(double dx, double dy) {
     }
 }
 
+void performTrackpadScroll(double dx, double dy) {
+    // Windows 10 has no public precision-touchpad injection API. Replay each
+    // native trackpad frame as an ordinary high-resolution wheel event while
+    // preserving the canonical 1x magnitude.
+    const int horizontal_distance =
+        logicalTrackpadScrollToWindowsWheel(dx, false);
+    const int vertical_distance =
+        logicalTrackpadScrollToWindowsWheel(dy, true);
+
+    if (horizontal_distance != 0) {
+        hscroll(horizontal_distance);
+    }
+    if (vertical_distance != 0) {
+        scroll(vertical_distance);
+    }
+}
+
 std::wstring stringToWstring(const std::string& str) {
     int wideCharLen = MultiByteToWideChar(CP_UTF8, 0, str.c_str(), -1, nullptr, 0);
     if (wideCharLen <= 0) return L"";
@@ -1464,6 +1497,13 @@ void HardwareSimulatorPlugin::HandleMethodCall(
         auto dx = (args->find(flutter::EncodableValue("dx")))->second;
         auto dy = (args->find(flutter::EncodableValue("dy")))->second;
         performMouseScroll(
+            std::get<double>(dx),
+            std::get<double>(dy));
+        result->Success(nullptr);
+  } else if (method_call.method_name().compare("trackpadScroll") == 0) {
+        auto dx = (args->find(flutter::EncodableValue("dx")))->second;
+        auto dy = (args->find(flutter::EncodableValue("dy")))->second;
+        performTrackpadScroll(
             std::get<double>(dx),
             std::get<double>(dy));
         result->Success(nullptr);
--- a/windows/hardware_simulator_plugin.cpp
+++ b/windows/hardware_simulator_plugin.cpp
@@ -4,6 +4,7 @@
 #include "desktop_service_input_client.h"
 #include "gamecontroller_manager.h"
 #include "notification_window.h"
+#include "trackpad_scroll_accumulator.h"
 #include "virtual_display_control.h"
 #include "SmartKeyboardBlocker.h"
 
@@ -55,6 +56,11 @@ constexpr UINT_PTR kCursorReseedSubclassId =
 // converts the logical page direction and clamps it to a safe Win32 range.
 constexpr double kMaxWindowsWheelDistance =
     static_cast<double>(WHEEL_DELTA) * 100.0;
+// Chromium's legacy Windows wheel path maps one 120-unit detent to 100 CSS
+// pixels with the default three-line setting. Convert native macOS trackpad
+// pixels into that Win32 unit space so cumulative page movement stays 1:1.
+constexpr double kWindowsTrackpadWheelUnitsPerPixel =
+    static_cast<double>(WHEEL_DELTA) / 100.0;
 
 int logicalScrollToWindowsWheel(double logical_distance, bool invert) {
   if (!std::isfinite(logical_distance)) {
@@ -65,22 +71,6 @@ int logicalScrollToWindowsWheel(double logical_distance, bool invert) {
   const double clamped = (std::clamp)(
       scaled, -kMaxWindowsWheelDistance, kMaxWindowsWheelDistance);
   return static_cast<int>(std::lround(clamped));
-}
-
-int logicalTrackpadScrollToWindowsWheel(
-    double logical_distance,
-    bool invert) {
-  if (!std::isfinite(logical_distance)) {
-    return 0;
-  }
-  const double direction = invert ? -1.0 : 1.0;
-  const double converted = logical_distance * direction;
-  const double clamped = (std::clamp)(
-      converted, -kMaxWindowsWheelDistance, kMaxWindowsWheelDistance);
-  // Win32 MOUSEINPUT::mouseData is integral. The trackpad path deliberately
-  // sends the canonical 1x delta directly at that API boundary: no mouse-wheel
-  // sensitivity multiplier, fractional accumulator, or nearest-value rounding.
-  return static_cast<int>(clamped);
 }
 
 LRESULT CALLBACK CursorReseedSubclassProc(
@@ -1413,12 +1403,17 @@ void performMouseScroll(double dx, double dy) {
 
 void performTrackpadScroll(double dx, double dy) {
     // Windows 10 has no public precision-touchpad injection API. Replay each
-    // native trackpad frame as an ordinary high-resolution wheel event while
-    // preserving the canonical 1x magnitude.
+    // native trackpad frame as an ordinary high-resolution wheel event. The
+    // legacy API is integral, so each axis carries its fractional wheel-unit
+    // remainder forward to preserve cumulative page distance.
+    thread_local TrackpadScrollAccumulator horizontal_accumulator(
+        kWindowsTrackpadWheelUnitsPerPixel, kMaxWindowsWheelDistance);
+    thread_local TrackpadScrollAccumulator vertical_accumulator(
+        kWindowsTrackpadWheelUnitsPerPixel, kMaxWindowsWheelDistance);
     const int horizontal_distance =
-        logicalTrackpadScrollToWindowsWheel(dx, false);
+        horizontal_accumulator.Convert(dx, false);
     const int vertical_distance =
-        logicalTrackpadScrollToWindowsWheel(dy, true);
+        vertical_accumulator.Convert(dy, true);
 
     if (horizontal_distance != 0) {
         hscroll(horizontal_distance);
--- a/example/macos/RunnerTests/RunnerTests.swift
+++ b/example/macos/RunnerTests/RunnerTests.swift
@@ -24,4 +24,13 @@ class RunnerTests: XCTestCase {
     waitForExpectations(timeout: 1)
   }
 
+  func testTrackpadPointDeltaUsesLogicalInverseAtOneToOneScale() {
+    XCTAssertEqual(HardwareSimulatorPlugin.nativeTrackpadPointDelta(-30), 30)
+    XCTAssertEqual(HardwareSimulatorPlugin.nativeTrackpadPointDelta(30), -30)
+    XCTAssertEqual(HardwareSimulatorPlugin.nativeTrackpadPointDelta(-0.1), 1)
+    XCTAssertEqual(HardwareSimulatorPlugin.nativeTrackpadPointDelta(0.1), -1)
+    XCTAssertEqual(HardwareSimulatorPlugin.nativeTrackpadPointDelta(0), 0)
+    XCTAssertNil(HardwareSimulatorPlugin.nativeTrackpadPointDelta(.nan))
+  }
+
 }
--- a/lib/hardware_simulator.dart
+++ b/lib/hardware_simulator.dart
@@ -1,6 +1,10 @@
 import 'hardware_simulator_platform_interface.dart';
 export 'hardware_simulator_platform_interface.dart'
-    show DisplayCountChangedCallback;
+    show
+        DisplayCountChangedCallback,
+        TrackpadScrollCallback,
+        TrackpadScrollEvent,
+        TrackpadScrollPhase;
 import 'display_data.dart';
 
 /// Snapshot of the macOS TCC permissions relevant to remote-control hosting.
@@ -224,6 +228,16 @@ class HardwareSimulator {
 
   static void removeCursorWheel(CursorWheelCallback callback) {
     HardwareSimulatorPlatform.instance.removeCursorWheel(callback);
+  }
+
+  /// Registers for precision trackpad scrolling captured by the native
+  /// platform embedder.
+  static void addTrackpadScroll(TrackpadScrollCallback callback) {
+    HardwareSimulatorPlatform.instance.addTrackpadScroll(callback);
+  }
+
+  static void removeTrackpadScroll(TrackpadScrollCallback callback) {
+    HardwareSimulatorPlatform.instance.removeTrackpadScroll(callback);
   }
 
   // ignore: constant_identifier_names
--- a/lib/hardware_simulator.dart
+++ b/lib/hardware_simulator.dart
@@ -73,6 +73,25 @@ class HWMouse {
   void performMouseScroll(double dx, double dy) {
     HardwareSimulatorPlatform.instance.performMouseScroll(dx, dy);
   }
+
+  /// Injects a continuous precision-trackpad delta without discarding
+  /// fractional pixels.
+  ///
+  /// Positive x scrolls right and positive y scrolls down. Platforms without
+  /// a native precise path may fall back to ordinary scroll injection.
+  void performTrackpadScroll(
+    double dx,
+    double dy, {
+    TrackpadScrollPhase phase = TrackpadScrollPhase.none,
+    bool isMomentum = false,
+  }) {
+    HardwareSimulatorPlatform.instance.performTrackpadScroll(
+      dx,
+      dy,
+      phase: phase,
+      isMomentum: isMomentum,
+    );
+  }
 }
 
 class GameController {
--- a/lib/hardware_simulator_platform_interface.dart
+++ b/lib/hardware_simulator_platform_interface.dart
@@ -33,7 +33,9 @@ enum TrackpadScrollPhase {
 ///
 /// [x] and [y] use Flutter view coordinates. Deltas use page direction:
 /// positive x scrolls right and positive y scrolls down. Every platform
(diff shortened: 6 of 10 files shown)


zhuhaichao518:
@hxyshadow 麻烦 review 一下这笔统一原生触控板事件接口，重点看 macOS NSEvent 精确滚动筛选、页面方向归一化、phase/momentum 元数据，以及首个监听启动/末个监听停止的生命周期。它是 cloudplayplus#442 的依赖。

macos/Classes/HardwareSimulatorPlugin.swift
```diff
@@ -73,6 +77,58 @@ public class HardwareSimulatorPlugin: NSObject, FlutterPlugin {
 #endif
   }
 
+  private func startTrackpadScrollCapture() {
+    guard trackpadScrollMonitor == nil else { return }
+    trackpadScrollMonitor = NSEvent.addLocalMonitorForEvents(
+      matching: .scrollWheel
+    ) { [weak self] event in
+      self?.handleTrackpadScroll(event)
+      return event
```
hxyshadow, line 91: 💡 **非消费式捕获 → 消费端要注意双源。** 这里 `return event` 是有意的(不吞掉 AppKit 事件,避免全局破坏 app 自身 UI 滚动),这个设计是对的。但副作用是:同一次物理滚动,Flutter 仍会照常派发它自己的 `PointerPanZoom`/`PointerScroll`。所以消费端(cloudplayplus#442)在启用这条原生轨迹后,必须**主动忽略** Flutter 自带的 macOS 滚动源,否则会双倍滚动。非阻塞,只是提醒把这个契约写进消费端 PR。

hxyshadow (commented):
**💬 COMMENT(评论)**

## Code Review: #41 — feat(macos): expose native trackpad scroll events

**Verdict: Approved ✅**（0 阻塞，1 条非阻塞提示）
**Author:** @zhuhaichao518 | **Files:** 5 (+258 -2) | 首轮 review（无 claude 前置轮）
**Reviewed against:** `1eee628cfa3eaeb160fc15cb1abd2ec0a55c883e`

统一原生触控板事件接口设计干净，重点看的四块都过了：

### ✅ 已核对（你点名的重点）
- **首听启动 / 末听停止的生命周期** — `addTrackpadScroll` 在 add 之前先取 `shouldStartCapture = isEmpty`，`removeTrackpadScroll` 只在 `removed && isEmpty` 时 stop；native 侧 `start/stopTrackpadScrollCapture` 都是幂等的（`guard trackpadScrollMonitor == nil` / `guard let trackpadScrollMonitor`）；`deinit` 里还兜底调了一次 `stopTrackpadScrollCapture()`。对称、无泄漏、无重复 add monitor。
- **NSEvent 精确滚动筛选** — `hasPreciseScrollingDeltas` + `event.window === flutterView.window` + `eventPhase.rawValue != 0` 三重过滤，能把普通鼠标滚轮/其它窗口的事件挡在外面，只放行本窗口的真实触控板手势。
- **页面方向归一化** — `dx/dy = -event.scrollingDelta*`，注释也点明了「AppKit 报手势方向、API 用页面方向」；坐标系用 `isFlipped` 正确翻转 Y。契约（+x 向右、+y 向下）在 `TrackpadScrollEvent` doc 里写清楚了。
- **phase / momentum 元数据** — `isMomentum = momentumPhase.rawValue != 0`，momentum 阶段优先取 `momentumPhase` 否则取 `phase`，`trackpadScrollPhaseName` 按 cancelled→ended→began→mayBegin→stationary→changed 的优先级映射，Dart 侧 `_trackpadScrollPhase` 对未知值兜底到 `.none`，`_asDouble` 对非 num 兜底到 0——解码健壮。

### ✅ 跨文件 & 平台
- 导出在 `hardware_simulator.dart` 补齐（`TrackpadScrollCallback/Event/Phase`）。
- 平台接口默认实现是**安全 no-op**（`addTrackpadScroll` 空体、`startTrackpadScrollCapture` 返回 false），刻意区别于其它 `throw UnimplementedError` 的兄弟方法——符合「未实现平台安全降级」的 PR 意图，Web 直接继承 no-op，不会双发。
- 监听闭包用 `[weak self]`，无循环引用；`deinit` 仍能触发并 `removeMonitor`。

### ✅ 测试
两个用例分别锁住了「首/末监听驱动 native 启停」和「onTrackpadScroll 解码」，覆盖到位。

### 💡 建议（非阻塞，见 inline）
- **非消费式捕获的双源契约**：native monitor `return event`（不吞事件）是对的，但同一次物理滚动 Flutter 仍会发自己的 `PointerPanZoom`。消费端（cloudplayplus#442）启用这条原生轨迹时要主动忽略 Flutter 自带的 macOS 滚动源，否则双倍滚动。建议把这个契约写进消费端 PR。

draft 状态，方向和实现都没问题，先 APPROVE。

---

## Overview

Adds a cross-platform `TrackpadScrollEvent` API (position, page-direction deltas, gesture phase, momentum flag) with reference-counted native capture start/stop, a macOS implementation via `NSEvent.addLocalMonitorForEvents(matching: .scrollWheel)`, and no-op defaults elsewhere. 258/-2 across 5 files, with two unit tests.

Overall this is well-shaped: the platform-interface defaults are no-ops rather than `UnimplementedError` (right choice for an additive, incrementally-implemented feature), the monitor returns the event so AppKit delivery isn't consumed, and the flipped/unflipped Y conversion is handled explicitly.

## Issues & suggestions

**1. `registrar.view` is captured once at registration (`HardwareSimulatorPlugin.swift:62`)** — the strongest concern. `FlutterPluginRegistrar.view` is `NSView?` and can be nil at plugin-registration time depending on when the registrant runs relative to `FlutterViewController` view creation. Combined with `private weak var flutterView`, a nil-at-registration (or later-recreated) view means `handleTrackpadScroll` silently `guard`s out forever and the feature is dead with no diagnostic. Suggest resolving the view lazily from the registrar at capture-start time, and/or falling back to `NSApp.keyWindow?.contentView`.

**2. `startTrackpadScrollCapture` always returns `true` on macOS (`:2364`)** — the interface documents the return as "false on platforms that have not implemented the unified source," but macOS returns true even when `flutterView == nil`, i.e. when no event will ever be delivered. Returning `flutterView != nil` would make the contract honest.

**3. Concurrent modification when a listener removes itself.** `for (var callback in trackpadScrollCallbacks) callback(event)` throws `ConcurrentModificationError` if a callback calls `removeTrackpadScroll` on itself. This is a very plausible pattern for this API specifically (deregister on `ended`/`cancelled` phase), more so than for the existing wheel callbacks. Iterate a copy: `for (final cb in List.of(trackpadScrollCallbacks))`. Same for the ordering guarantee — `removed && isEmpty` is correct, but a duplicate-add makes the refcount wrong; consider guarding against duplicate registration.

**4. Natural-scrolling direction.** `dx/dy` are a hardcoded negation of `scrollingDeltaX/Y`, but AppKit already applies the user's "natural scrolling" preference to those deltas. So the documented "positive y scrolls down" contract inverts depending on a system setting. If following the user's preference is intentional, say so in the doc comment; if the contract is meant to be absolute, consult `event.isDirectionInvertedFromDevice`.

**5. Magic Mouse / non-trackpad devices.** `hasPreciseScrollingDeltas` is true for Magic Mouse too, so the "trackpad" naming is slightly broader than reality. Not a bug, but worth a doc note.

**6. Multiple registrations.** If the plugin is registered against more than one registrar (multi-window / multi-engine), each instance installs its own local monitor and each will fire for the matching window — mostly self-limiting via the window check, but worth being aware of.

## Tests

Good coverage of the two things worth testing (refcounted lifecycle, decode path). Two nits:
- `test('decodes native trackpad scroll events')` never removes its listener, so it leaves capture "started" — harmless if `platform` is rebuilt in `setUp`, but worth confirming.
- No test for the unknown-phase → `TrackpadScrollPhase.none` fallback or the non-`Map` arguments guard, both of which are defensive branches you deliberately wrote.

## Minor

- `removeTrackpadScroll` and `stopTrackpadScrollCapture` in the public API / interface lack doc comments while their `add`/`start` counterparts have them.
- No CHANGELOG entry or version bump for a public API addition — check whether this repo requires one.
- The `ensureConsoleForDisplay` reformat is unrelated formatting noise; fine to leave.

## Risk

Low blast radius — purely additive, no behavior change for existing callers, and the monitor returns the event unconsumed so it can't break normal Flutter scroll handling. No security concerns; the local monitor only sees this app's events.

zhuhaichao518:
已在 `1209008` 处理 review 建议：capture 启动时从 registrar 懒加载 Flutter view、view 不可用时返回 false，窗口/view 变化时重新解析；回调改为副本遍历并去重注册，覆盖回调自注销；补充系统滚动方向契约和 API 注释。插件 6 项测试及父应用 macOS release build 均通过，麻烦重新确认。

hxyshadow (commented):
**💬 COMMENT(评论)**

## 🤖 Hermes reviewer — **APPROVE**（re-review）

## Code Review: #41 — feat(macos): expose native trackpad scroll events

**Verdict: Approved ✅**（0 阻塞）
**Author:** @zhuhaichao518 | **Files:** 5 (+280 -2)
**Reviewed against:** `1209008be8f070164b8cd6754f93da26ff6f1bfc`（上轮 `1eee628` 之后的 fix 提交）

上轮我 + Claude 提的 4 条实质问题，逐条在 `1209008` 落地核对，全部 ✅：

### ✅ Carry-forward 核对（prior 4 项）
1. **`registrar.view` 注册期捕获可能为 nil → 特性静默失效**（Claude #1，最强的一条）— 已改：新增 `private weak var registrar`，`register` 里改存 registrar 而非立刻取 view；`startTrackpadScrollCapture` 在启动时才 `registrar?.view ?? registrar?.viewController?.view` 懒解析；`handleTrackpadScroll` 里 window 不匹配还会**重新解析** view。彻底消除「注册期 view 未就绪 → 永久 guard out」。✅
2. **`startTrackpadScrollCapture` 恒返回 true 与接口契约不符**（Claude #2）— 已改：`if monitor != nil { return flutterView != nil }`，新建路径 `guard flutterView != nil else { return false }`，MethodChannel 直接 `result(startTrackpadScrollCapture())`。view 不可用时诚实返回 false。✅
3. **回调自注销引发 `ConcurrentModificationError` + 重复注册破坏 refcount**（Claude #3）— 已改：派发改为 `for (final callback in List<TrackpadScrollCallback>.of(trackpadScrollCallbacks))` 副本遍历；`addTrackpadScroll` 加 `if (contains(callback)) return` 去重。✅
4. **自然滚动方向随系统设置反转，契约未说明**（Claude #4）— 已改：`TrackpadScrollEvent` doc 补了「resulting page direction follows the user's platform scroll-direction preference」。契约明确。✅
   - 附带：`removeTrackpadScroll` / `stopTrackpadScrollCapture` 的 doc 注释也补齐了（Claude Minor）。

### ✅ 新增测试核对（本轮新 commit）
- lifecycle 用例新增**重复 add 同一 listener**行（`addTrackpadScroll(firstListener)` ×2），锁住去重不破坏 refcount。
- decode 用例改为**回调内自注销**（`listener` 内调 `removeTrackpadScroll(listener)`）+ 尾部 `await Future.delayed(zero)`，正好回归覆盖了 #3 的并发遍历路径。测试设计对准了新修的两个点。✅

### 💡 沿用（非阻塞，上轮已提，仍适用）
- **非消费式捕获的双源契约**：native monitor `return event`（不吞 AppKit 事件）是对的，但同一次物理滚动 Flutter 仍会派发自己的 `PointerPanZoom`。消费端 cloudplayplus#442 启用这条原生轨迹时要主动忽略 Flutter 自带 macOS 滚动源，否则双倍滚动——建议写进消费端 PR。

修复干净、对称、有测试兜底。draft 状态，方向与实现都稳，维持 APPROVE。

---

## Overview

Adds a cross-platform `TrackpadScrollEvent` API (position in Flutter view coords, page-direction deltas, gesture phase, momentum flag), with reference-counted native capture start/stop, a macOS implementation via a *local* `NSEvent` scrollWheel monitor, and no-op defaults elsewhere. 280/-2 across 5 files, 2 new tests.

## What's good

- Local monitor returns the event unchanged — AppKit event flow is not consumed, so Flutter's own pointer pipeline still works.
- `[weak self]` in the monitor block, `deinit` removes the monitor.
- Dispatch iterates `List.of(trackpadScrollCallbacks)` — safe against a listener removing itself mid-dispatch (the second test exercises exactly this).
- `MissingPluginException` swallowed so unimplemented platforms degrade to silence rather than crashing.
- Base-class methods are no-op/`false` defaults rather than `UnimplementedError`, consistent with the "incremental platform impls" goal (though it diverges from the `throw UnimplementedError` style used by every other method in the file — intentional, but worth a one-line comment).

## Issues / suggestions

**1. `startTrackpadScrollCapture()`'s return value is discarded (`hardware_simulator_method_channel.dart`)**
`addTrackpadScroll` calls `unawaited(startTrackpadScrollCapture())`. If native returns `false` (no Flutter view yet, or unsupported platform), the listener is registered and simply never fires, with no signal to the caller. Consider having `addTrackpadScroll` return `Future<bool>`, or log at least once on failure. The macOS `guard flutterView != nil else { return false }` path is genuinely reachable at early startup, so this isn't hypothetical.

**2. Sign convention doc is self-contradictory (`hardware_simulator_platform_interface.dart`)**
The doc says deltas are canonical ("positive y scrolls down… every platform normalizes to this contract") and then says the direction "follows the user's platform scroll-direction preference." Those can't both hold — with natural scrolling on/off, a raw negation of `scrollingDeltaY` flips meaning. Either drop the canonical claim, or use `event.isDirectionInvertedFromDevice` to normalize for real. As written, downstream consumers can't tell which contract to code against.

**3. `stopTrackpadScrollCapture()` doesn't clear `flutterView` (macOS)**
Minor, but leaves stale view state across stop/start cycles. Set `flutterView = nil` alongside the monitor.

**4. Hot restart leaves the monitor running (macOS)**
Dart-side listeners are wiped on hot restart, but the plugin instance and its monitor survive; the native side keeps posting `onTrackpadScroll` into a channel nobody listens to. Not harmful, but a `detachFromEngine`/re-`register` hook that calls `stopTrackpadScrollCapture()` would keep it clean. Same reasoning applies if `register` is ever called twice.

**5. Add/remove start/stop race**
`add → remove` within the same microtask fires `start` then `stop`. Channel ordering is FIFO so it resolves correctly today, but the state isn't guarded by anything — a future `await` between the checks would break it. A simple `bool _captureRequested` guard would make the intent explicit.

**6. Test coverage gaps**
Covered: refcounting and happy-path decode. Not covered: non-`Map` arguments (the `if (arguments is Map)` branch), unknown/missing `phase` → `none`, non-numeric fields → `_asDouble` returning `0`, and `MissingPluginException` handling. The `_asDouble` silent-zero fallback in particular deserves a test since it turns malformed native payloads into plausible-looking `(0,0)` events.

## Performance / security

- Precision-only filter (`hasPreciseScrollingDeltas`) plus the `phase.rawValue != 0` guard keeps channel traffic bounded to real gestures; per-event `invokeMethod` on a high-frequency trackpad stream is still ~120 msgs/sec, acceptable but worth knowing.
- Local (not global) monitor means no accessibility/input-monitoring permission is required and no other app's events are observed — the right choice here, no added privacy surface.

Nothing blocking; the items above are correctness-adjacent polish rather than defects that would break the happy path.

zhuhaichao518:
复审已通过。另在 `74c18bb` 顺手处理了新增检查中的低风险生命周期建议：停止捕获时无论 monitor 是否存在都会清空缓存的 Flutter view，避免 stop/start 周期保留过期引用。插件 6 项测试与父应用 macOS release build 均通过。API 启动失败反馈、方向契约和 Magic Mouse 双源涉及行为设计，保持现状等待实测/决策。

hxyshadow (approved):
**✅ APPROVE(通过)**

## 🤖 Hermes reviewer — **APPROVE**（re-review · A2 静默 push）

## Code Review: #41 — feat(macos): expose native trackpad scroll events

**Verdict: Approved ✅**（0 阻塞）
**Author:** @zhuhaichao518 | **Files:** 5 (+282 -2) | draft
**Reviewed against:** `74c18bb4d62f71feca3881de4c4e69cae1161986`（上轮 `1209008` 之后新增 1 个 fix commit `74c18bb`）

本轮增量只有一个 commit `fix(macos): clear trackpad capture view on stop`，只改 `HardwareSimulatorPlugin.swift`（+4/-2）。逐行核对干净：

### ✅ 本轮新增 commit 核对
- **`stopTrackpadScrollCapture` 停止时清空缓存 view**（对应 Claude 上轮 R2 #3）— 已改：`guard let trackpadScrollMonitor else { return }` 换成 `if let trackpadScrollMonitor { removeMonitor(...) }`，从而 `flutterView = nil` **无条件执行**（旧写法在 monitor 已为 nil 时会提前 return，漏清 view）。彻底消除 stop→start 周期里保留过期 `flutterView` 引用的问题。✅
- **无回归验证**：清空 view 是安全的，因为两条使用路径都会重新惰性解析——
  - `startTrackpadScrollCapture` 走 `flutterView = registrar?.view ?? registrar?.viewController?.view`（:85），下次启动拿到全新 view；
  - `handleTrackpadScroll` 在 `flutterView?.window !== event.window` 时也会重解析（:107）。
  两者都不依赖 stop 后残留的 view，所以清空不会让下一轮捕获失效。✅
- **幂等性保持**：`deinit`→`stopTrackpadScrollCapture()`（:72）仍安全，monitor/view 双清；重复调用无副作用。✅

### 💡 沿用（非阻塞，作者已显式 defer 到实测/决策）
作者在 `74c18bb` 评论里明确保留三条行为设计项，等待实测/决策，合理：
- **API 启动失败反馈**（Claude R2 #1）：`addTrackpadScroll` 里 `unawaited(startTrackpadScrollCapture())` 丢弃返回值，早启动 view 未就绪时 listener 静默不触发、无信号。属可后补的观测性/契约改进，非阻塞。
- **方向契约文档自相矛盾**（Claude R2 #2）：canonical「+y 向下」与「跟随系统 scroll-direction 偏好」不能同时成立，需二选一或用 `isDirectionInvertedFromDevice` 真归一化。文档层问题，等消费端实测方向后定夺。
- **Magic Mouse / 双源契约**：`hasPreciseScrollingDeltas` 对 Magic Mouse 也为真；且非消费式 `return event` 意味着同一物理滚动 Flutter 仍派发自己的 `PointerPanZoom`。消费端 cloudplayplus#442 启用原生轨迹时须主动忽略 Flutter 自带 macOS 滚动源，否则双倍滚动——建议写进消费端 PR。

修复干净、对称、动机清晰，无回归。维持 APPROVE。

---

## Overview

Adds a cross-platform `TrackpadScrollEvent` contract (view-space position, page-direction deltas, gesture phase, momentum flag), ref-counted native capture start/stop driven by listener registration, and a macOS implementation via a local `NSEvent` monitor on `.scrollWheel`. Other platforms are no-ops via non-throwing defaults on the platform interface.

## What's good

- **Non-throwing defaults** on `HardwareSimulatorPlatform` for the four new members (rather than `UnimplementedError` like the rest of the interface) is the right call here — it keeps unimplemented platforms as silent no-ops instead of crashing Viewer code. Deliberate deviation from convention, and correct.
- **The local monitor returns `event`**, so AppKit delivery to Flutter is not consumed. Good — the plugin observes rather than intercepts.
- **`MissingPluginException` is caught** on both invoke sites, so a stale native side degrades instead of throwing into an unawaited future.
- **Callback list is copied** (`List.of(...)`) before dispatch, so a listener that removes itself during the callback can't corrupt iteration. The second test exercises exactly this.
- Sign inversion and the gesture-vs-page-direction rationale are commented at the call site and documented on the Dart class. `isFlipped` is handled rather than assumed.
- Tests cover the interesting behaviors: dedup on add, start-on-first / stop-on-last, and decode round-trip.

## Findings (all non-blocking)

**1. `startTrackpadScrollCapture()`'s result is discarded — silent failure.** `hardware_simulator_method_channel.dart:~270` fires it under `unawaited`, so if the native side returns `false` (macOS: `registrar?.view` is nil, which can happen if capture is started before the view is attached), listeners stay registered and simply never receive events, with no signal. Consider either surfacing this (make `addTrackpadScroll` return `Future<bool>`, or log in debug) or retrying view resolution lazily — the macOS `handleTrackpadScroll` already re-resolves the view, so the only thing blocking recovery is that the monitor was never installed.

**2. Stopping mid-gesture leaves consumers without a terminal phase.** If the last listener is removed while a gesture or momentum train is in flight, the monitor is torn down and no `ended`/`cancelled` ever arrives. Any consumer holding per-gesture state keyed on `began` would leak that state. Worth a line in the `TrackpadScrollEvent` doc comment saying a terminal phase is not guaranteed.

**3. `@immutable` without value semantics.** `TrackpadScrollEvent` has no `==`/`hashCode`/`toString`. `toString` in particular would help a lot when debugging scroll-direction issues, which is exactly the class of bug this feature exists to fix.

**4. `_asDouble` coerces missing/garbage to `0` silently.** Fine as defensive coding since the sender is first-party, but it means a native-side key rename shows up as "scrolling stopped working" rather than an error. Same for `_trackpadScrollPhase`'s `_ => none` fallback.

**5. macOS: per-event re-resolution on foreign windows.** In `handleTrackpadScroll`, when the event comes from a different window the `flutterView?.window !== event.window` check re-reads `registrar?.view` on *every* such event before failing the guard below. Trackpad scroll events are high-frequency; this is cheap but pointless. Caching the resolution failure, or only re-resolving when `flutterView == nil`, would avoid it.

**6. `deinit` cleanup is mostly theoretical.** Flutter retains the method-call delegate for the engine's lifetime, so `deinit` rarely runs. Not a leak (the explicit `stopTrackpadScrollCapture` channel method is the real path), just noting the safety net isn't load-bearing.

**7. Test gap:** no coverage of the `MissingPluginException` path or of a malformed/non-`Map` `onTrackpadScroll` payload, both of which the code explicitly handles. Cheap to add.

## Correctness / security / performance

No blocking correctness issue found. Momentum detection (`momentumPhase.rawValue != 0`, with `event.phase` being `.none` during momentum) is sound, and the `hasPreciseScrollingDeltas` guard correctly excludes physical wheels. `NSEvent.addLocalMonitorForEvents` is app-local — no extra entitlement, no system-wide input observation, so no new privacy surface. The unrelated `ensureConsoleForDisplay` reformat is harmless formatter noise.

zhuhaichao518:
@hxyshadow 实测发现 macOS Host 复用了整数滚轮注入，原生 `0.1` 会在 `rounded()` 后变成 0。已在 `8dfaa2b` 增加统一 precise 注入：保留 Double 小数，重建 wheel/point + 16.16 fixed-point，设置 continuous，并通过公开的 CoreGraphics scroll/momentum phase 字段注入生命周期；普通 mouseScroll 不变。插件 7 项测试、父应用 78 项定向测试、analyze 和 macOS release build 均通过，麻烦基于最新提交复审。

lib/hardware_simulator_method_channel.dart
```diff
@@ -375,6 +433,37 @@ class MethodChannelHardwareSimulator extends HardwareSimulatorPlatform {
...
   }
 
+  @override
+  Future<void> performTrackpadScroll(
+    double dx,
+    double dy, {
+    TrackpadScrollPhase phase = TrackpadScrollPhase.none,
+    bool isMomentum = false,
+  }) async {
+    await methodChannel.invokeMethod('trackpadScroll', {
```
hxyshadow: 💡 **文档承诺的「fallback 到普通滚动」在这条路径上不会发生。** facade 注释（`hardware_simulator.dart:80-81`）写的是「Platforms without a native precise path may fall back to ordinary scroll injection」。抽象基类的默认实现（`hardware_simulator_platform_interface.dart:285` `return performMouseScroll(dx, dy)`）确实是这个语义，但运行期的真实实例是 `MethodChannelHardwareSimulator`，它 `@override` 掉了默认实现，无条件 `invokeMethod('trackpadScroll')`。Windows 原生 `switch` 没有 `trackpadScroll` case → `result->NotImplemented()`（`windows/hardware_simulator_plugin.cpp:1942`）→ Dart 侧抛 `MissingPluginException`，而这里是 `await` 且未捕获，会直接冒泡给调用方，并不会退回 `performMouseScroll`。

本 PR 是 macOS-only，消费端 cloudplayplus#442 也按 macOS 门控，所以现在触发不到，非阻塞。但如果想让 API 契约名副其实，建议二选一：（a）在这个 override 里 `try { invokeMethod('trackpadScroll') } on MissingPluginException { performMouseScroll(dx, dy) }`，真正实现声明的 graceful fallback；（b）或者把 facade 注释改成「非 macOS 平台暂未实现，会抛 MissingPluginException」，避免误导后续接入者。

hxyshadow (changes requested):
**🔴 REQUEST_CHANGES(需修改)**

## Code Review: #41 — feat(macos): expose native trackpad scroll events（复审 · A2/Case-A 增量）

**Verdict: Approved ✅**（0 阻塞，1 条非阻塞 💡 · inline）
**Author:** @zhuhaichao518 | **Files:** 5 (+479 -2) | draft
**Reviewed against:** `8dfaa2b`（上轮 APPROVE 针对 `74c18bb`，作者随后推 1 个 commit `8dfaa2b feat(macos): inject precise trackpad scroll events`，并 @我基于最新提交复审）

上轮我 + Claude 的实质问题已在 `74c18bb` 前全部落地并 APPROVE。本轮增量只有一个 commit —— 补 macOS **Host 端**精确注入路径（对应「实测发现整数滚轮注入把原生 `0.1` rounded() 成 0」）。作为迷你 PR 走了一遍 Step 4 + HE 系列（host 输入注入 shim）角度。

### ✅ 新 commit `8dfaa2b` 复查

- **保真注入的核心思路正确**（`HardwareSimulatorPlugin.swift:1986-2044`）：用 `CGEvent(scrollWheelEvent2Source:units:.pixel:...)` 重建连续事件，`scrollWheelEventFixedPtDeltaAxis1/2` 写 16.16 fixed-point 保留小数，`scrollWheelEventPointDeltaAxis1/2` 写整数轴，`IsContinuous=1`，`ScrollPhase`/`MomentumPhase` 按 `isMomentum` 二选一注入——普通 `mouseScroll` 路径不动。等价于捕获侧 `NSEvent` 的逆过程。
- **符号约定自洽**：捕获侧 `dx/dy = -event.scrollingDeltaX/Y`（gesture→page，`swift:125-126`）；注入侧 fixedPt 字段写 `-dy/-dx`（page→gesture），point 轴保留 page 方向，与注释「wheel/point 轴保留 page direction，16.16 字段用相反的 gesture-direction」一致。round-trip 方向不翻。
- **`nativeTrackpadPointDelta`**（`swift:1938-1953`）：`isFinite` 守卫 + `0→0`；`(logicalDelta*10).rounded()` 后若归零则夹到 ±1，保证亚像素输入不被静默吞掉；`Int32.min/max` 夹紧防溢出。健壮。
- **协议对称性**：新 method-channel case `trackpadScroll`（`swift:2755-2770`）参数缺失走 `FlutterError(BAD_ARGS)`，与既有 `mouseScroll` 分支对称；Dart 侧新增单测 `performTrackpadScroll preserves fractional deltas` 断言 `0.125 / -0.375 / phase / isMomentum` 原样透传（fractional 不丢）。

### 💡 1 条非阻塞（inline · `hardware_simulator_method_channel.dart:443`）

facade 注释承诺「无原生精确路径的平台可 fallback 到普通滚动」，但真实实例 `MethodChannelHardwareSimulator` `@override` 了基类默认的 `performMouseScroll` fallback，无条件 `invokeMethod('trackpadScroll')`。Windows 原生无该 case → `NotImplemented()` → `MissingPluginException` 冒泡，并不会退回普通滚动。本 PR macOS-only、消费端也按 macOS 门控，触发不到，非阻塞。建议要么在 override 里 `on MissingPluginException` catch 回退，要么把注释改成「非 macOS 未实现」。详见 inline。

### 💡 沿用（上轮已提，仍适用，非阻塞）

- **非消费式捕获的双源契约**（`swift:91` `return event`）：native monitor 不吞 AppKit 事件是对的，但同一次物理滚动 Flutter 仍派发自己的 `PointerPanZoom`。消费端 cloudplayplus#442 启用原生轨迹时须主动忽略 Flutter 自带 macOS 滚动源，否则双倍滚动——建议写进消费端 PR。
- 方向契约里「canonical page direction」与「follows platform scroll-direction preference」的措辞张力、启动失败反馈——作者已说明留待实测/决策，保持 ☑️。

### ✅ Looks Good

- 保真注入用重建 CGEvent + fixed-point 而非放大取整，选型稳；亚像素兜底到 ±1 避免静默吞。
- 增量自带单测 + 父应用 78 项定向测试 / analyze / macOS release build 全过。

无回归，无 critical/warning 级残留。维持 APPROVE。

---

## 🧠 Claude review — **REQUEST_CHANGES**

## Overview

Adds a cross‑platform precision‑trackpad contract: `TrackpadScrollEvent`/`TrackpadScrollPhase`, listener add/remove with refcounted native capture start/stop, and `performTrackpadScroll` injection. macOS implements capture via a local `NSEvent` monitor for `.scrollWheel` (non‑consuming) and injection via `CGEvent(scrollWheelEvent2Source:)` with 16.16 fixed‑point deltas plus scroll/momentum phase fields. Other platforms are safe no‑ops / fall back to `performMouseScroll`.

Design is clean and consistent with the existing plugin structure; the Dart side is well factored and reasonably tested.

## Blocking concerns

**1. `nativeTrackpadPointDelta` scales point deltas ×10 — likely a 10× over‑scroll** (`macos/Classes/HardwareSimulatorPlugin.swift:1938`)

```swift
var rounded = (logicalDelta * 10).rounded()
if rounded == 0 { rounded = logicalDelta.isLess(than: 0) ? -1 : 1 }
```

The comment justifies this from a single observation (`scrollingDelta=0.1` → point axis 1). But that data point is equally explained by macOS clamping sub‑pixel deltas to a minimum of 1 — which the `rounded == 0` branch already handles. On real macOS precise events, `scrollWheelEventPointDeltaAxis*` tracks `scrollingDelta` ~1:1. With this formula a normal `dy = 30` gesture injects a point delta of 300. Any consumer that reads point deltas rather than the fixed‑point fields (which is most non‑AppKit / Chromium‑style consumers) will scroll ~10× too far. Please verify against a dump of real events across magnitudes, not just 0.1, and drop the multiplier if it doesn't hold.

**2. Sign convention is inconsistent with the existing mouse‑scroll path** (same function)

`nativeWheelDelta` (line ~1910) negates: `(-logicalDelta).rounded()`, with the comment "RD_MOUSE_SCROLL uses logical page direction (+x right, +y down). CoreGraphics wheel1/wheel2 use physical direction". The new trackpad path does *not* negate the wheel/point axes and comments the opposite ("wheel/point axes retain page direction"). Both cannot be true — `wheel1`/`wheel2` in `scrollWheelEvent2Source` and `scrollWheelEventPointDeltaAxis1/2` are the same axis with the same sign. As written, injected trackpad scrolling is inverted relative to injected mouse scrolling. One of the two comments/implementations needs fixing; please confirm empirically which.

## Non‑blocking suggestions

- **`startTrackpadScrollCapture` result is discarded.** `addTrackpadScroll` calls `unawaited(startTrackpadScrollCapture())` and ignores the `false` return, so on an unsupported platform (or when `registrar?.view` is nil) the listener stays registered and silently never fires. Consider surfacing this — e.g. keep `addTrackpadScroll` but expose the future, or log.
- **Unhandled async errors.** Both `unawaited(...)` calls only swallow `MissingPluginException`. A `PlatformException` from native becomes an unhandled zone error. Add a `.catchError` or broaden the catch.
- **Start/stop race.** `add`/`remove` fire‑and‑forget; a fast remove‑then‑add sequence can let `stopTrackpadScrollCapture` land after `startTrackpadScrollCapture`, leaving a live listener with no capture. Serializing on a single `Future` chain would close this.
- **`flutterView` re‑resolution in `handleTrackpadScroll`** runs on every scroll event whose window doesn't match. That's cheap but noisy; a cached-and-invalidated-on-window-change approach reads better.
- **`@immutable` without `==`/`hashCode`** on `TrackpadScrollEvent` — harmless today, but add them (or use a record) if events ever land in sets/comparisons or widget tests.
- **`nativeTrackpadMomentumPhase("mayBegin"/"began")` → `.begin`** is fine, but `nativeTrackpadScrollPhase` maps `stationary` → `changed` while dropping `stationary` semantics; worth a one‑line comment since the Dart enum exposes `stationary` as distinct.
- Unrelated formatting fix in `ensureConsoleForDisplay` is fine to keep.

## Tests

Dart coverage is good for encoding, decoding, and listener refcounting. Nothing covers the two numeric conversions above — those live in Swift and are the risky part. A small Swift unit test (or at minimum a documented manual verification table of `logicalDelta` → observed native point/fixed‑point values) would materially de‑risk this.

## Security / performance

No new permissions or IPC surface beyond the existing channel. The local monitor returns the event unmodified, so AppKit delivery is preserved. Per‑event `invokeMethod` at trackpad rates (~120 Hz) is acceptable but is the highest‑frequency channel traffic in the plugin — worth watching if momentum sequences ever cause jank.

hxyshadow (commented):
**💬 COMMENT(评论)**

## 🤖 Hermes reviewer — **APPROVE**（复审 · A2 静默 push）

## Code Review: #41 — feat(macos): expose native trackpad scroll events

**Verdict: Approved ✅**（0 阻塞）
**Author:** @zhuhaichao518 | **Files:** 6 (+519 -2) | draft
**Reviewed against:** `33c9216`（上轮 CHANGES_REQUESTED 针对 `8dfaa2b`，作者随后推 1 个 commit `33c9216 fix(macos): preserve native trackpad scroll fields`）

上一轮 Claude 提了 2 条阻塞项 + 我提了 1 条 HE-7 inline。本轮增量只有这一个 fix commit，逐条在 `33c9216` 落地核对，全部 ✅。作为迷你 PR 走了一遍 Step 4 + HE 系列（host 注入 shim）角度，并跑了双视角门（语义/跨文件 + 测试/生命周期），无新增 critical/warning。

### ✅ Carry-forward 核对（上轮 3 项，全部落地）

1. **`nativeTrackpadPointDelta` ×10 疑似 10× 过冲**（Claude 阻塞 #1）— 已改（`HardwareSimulatorPlugin.swift:1938`）：`(logicalDelta * 10).rounded()` → `(-logicalDelta).rounded()`。整像素保留 1:1 幅度，只有非零亚像素才夹到 ±1。`dy=30` 现在注入 point delta = -30 而非 300，10× 过冲消除。✅
2. **符号约定与鼠标滚轮路径不一致 → 疑似注入反向**（Claude 阻塞 #2）— 已改：point 轴写 `verticalPoint/horizontalPoint`（已在 `nativeTrackpadPointDelta` 内负号取反），fixed-point 轴写 `-dy/-dx`，两个轴**一致取反**，与兄弟函数 `nativeWheelDelta`（`:1910` 同样 `(-logicalDelta).rounded()`）方向一致。无轴间反向，与既有鼠标滚轮注入方向自洽；`:2012` 注释也一并改正。✅
3. **facade 承诺的「fallback 到普通滚动」在真实实例上不生效**（我的 HE-7 inline `hardware_simulator_method_channel.dart:443`）— 已采纳方案 (a)：override 里 `try { invokeMethod('trackpadScroll') } on MissingPluginException { performMouseScroll(dx, dy) }`（`:440-449`）。声明的 graceful fallback 现在名副其实，非 macOS 平台不再直接冒泡 `MissingPluginException`。✅

### ✅ 新增测试核对（本轮 commit）

- **Swift `testTrackpadPointDeltaUsesLogicalInverseAtOneToOneScale`**：`-30→30 / 30→-30 / -0.1→1 / 0.1→-1 / 0→0 / .nan→nil`，六个断言手工对着实现逐条核过全对——正好锁住符号取反、1:1 幅度、亚像素夹紧、非有限值守卫这四点。为此把 `nativeTrackpadPointDelta` 从 private 提为 static，调用点同步改 `Self.`（`:1993-1994`），repo 内无其它调用者。
- **Dart `performTrackpadScroll falls back when native method is missing`**：断言调用顺序 `['trackpadScroll','mouseScroll']` 且 fallback 透传 `dx/dy`——直接回归覆盖我上轮 HE-7 inline 的路径。

### 💡 沿用（非阻塞，作者已显式 defer 到实测/决策）

- **非消费式捕获的双源契约**（`swift:91` `return event`）：native monitor 不吞 AppKit 事件是对的，但同一次物理滚动 Flutter 仍派发自己的 `PointerPanZoom`。消费端 cloudplayplus#442 启用原生轨迹时须主动忽略 Flutter 自带 macOS 滚动源，否则双倍滚动——建议写进消费端 PR。
- **方向契约措辞张力 / API 启动失败反馈**（Claude 上轮 #1、#2）：canonical「+y 向下」与「跟随系统 scroll-direction 偏好」措辞张力、`unawaited(startTrackpadScrollCapture())` 丢弃返回值——作者已说明留待实测/决策，保持 ☑️。

### ✅ Looks Good

- 保真注入用重建 CGEvent + fixed-point、point 轴 1:1、亚像素兜底 ±1，选型稳；两条阻塞的修法都带对应单测兜底。
- 增量自带 Swift + Dart 单测，父应用 78 项定向测试 / analyze / macOS release build 全过。

两条阻塞修复干净、对称、有测试兜底，无回归。上轮 CHANGES_REQUESTED 已清除，转 APPROVE。

---

## 概览

新增跨平台 trackpad 精确滚动能力：Dart 侧定义 `TrackpadScrollEvent` / `TrackpadScrollPhase` / 回调契约，method channel 实现监听器生命周期（首个监听器启动原生捕获、最后一个移除时停止）与注入 API；macOS 侧用 `NSEvent.addLocalMonitorForEvents(.scrollWheel)` 捕获（不消费事件），并用 `CGEvent(scrollWheelEvent2Source:)` + fixed-point 字段注入以保留小数像素。非实现平台为安全 no-op / 回退到 `performMouseScroll`。

整体设计清晰，方向约定（gesture → page direction）在两侧都有注释说明，测试覆盖了序列化、回退、生命周期、解码和符号转换，质量高于该仓库平均水平。

## 主要意见

**1. `startTrackpadScrollCapture` 失败后没有重试路径**（`HardwareSimulatorPlugin.swift:79`）
如果调用时 `registrar.view` 和 `viewController.view` 都还是 nil（引擎/视图尚未 attach），返回 `false`，Dart 侧只是把返回值丢掉——监听器仍在列表里，但原生捕获永远不会再启动，且 `removeTrackpadScroll` 之后 `stopTrackpadScrollCapture` 会照常发出。建议要么在 monitor 安装时不依赖 view（在 `handleTrackpadScroll` 里惰性解析，已有那段 re-resolve 逻辑），要么在 Dart 侧记录失败状态并在下次 add 时重试。

**2. 亚像素水平增量被放大到整点**（`nativeTrackpadPointDelta`）
把非零亚像素 clamp 到 ±1 对纵向合理（避免 0.1 被吃掉），但 point delta 字段是整数像素语义：60Hz 下持续的 0.05px 抖动会被放大成 60px/s 的可见漂移，尤其是横轴。只读 fixedPt 字段的应用没问题，读 point 字段的应用会漂。建议考虑累加余量（residual accumulator）而不是每次 clamp，或至少在注释里记下这个已知取舍。

**3. `startTrackpadScrollCapture` 在 monitor 已存在时的返回值**
`return flutterView != nil` — 若之前以 nil view 装过 monitor（当前代码不会，但 stop/start 交错时逻辑脆弱），返回 false 而 monitor 仍活着。让状态更单一（monitor 存在即 true）会更好读。

## 次要建议

- `TrackpadScrollEvent` 标了 `@immutable` 但没有 `==`/`hashCode`/`toString`，测试和调试时不太方便，建议补上。
- macOS 端 `trackpadScroll` 的参数解析用 `as? Double`：Dart 若传整数值（如 `dx: 0`），StandardCodec 会编成 Int，`as? Double` 失败 → 返回 BAD_ARGS。Dart 侧声明是 `double` 所以实际安全，但和同文件里 `mouseScroll` 的处理风格一致性值得确认；更稳妥是 `as? NSNumber`。
- `handleTrackpadScroll` 里 `flutterView?.window !== event.window` 时重新解析 view，每个非目标窗口的滚动事件都会走一次 registrar 查找；本地 monitor 频率很高（每帧多次），建议加个短路或只在 `flutterView == nil` 时解析。
- Dart 侧只有 trackpad 回调用了 `List.of(...)` 拷贝再遍历，其他回调没有——不是本 PR 的问题，但风格不一致，可在注释里说明原因（回调内会 remove 自身，测试里正是如此）。
- 请确认 `hardware_simulator_method_channel.dart` 已 import `dart:async`（`unawaited`）、interface 文件已 import `package:flutter/foundation.dart`（`@immutable`）；diff 里看不到，但既然 `flutter test` 通过应无问题。

## 风险 / 覆盖

- 安全面无新增风险：注入走已有的 `cghidEventTap` 路径，需要既有的 Accessibility 权限；捕获用 local monitor 且原样返回 event，不会吞掉 AppKit 事件。
- 测试：Dart 侧覆盖良好；Swift 侧只测了 `nativeTrackpadPointDelta`，phase 映射（`trackpadScrollPhaseName` / `nativeTrackpadScrollPhase` / `nativeTrackpadMomentumPhase`）三个纯函数同样容易测，建议补上——momentum/scroll phase 混淆是这类代码最典型的回归点。
- 无破坏性变更：平台接口新增方法都有默认实现，旧平台不受影响。

以上都不是必须阻塞合并的缺陷，第 1、2 点建议在合并前或紧随其后处理。

zhuhaichao518:
@hxyshadow 最新 `5d56475` 增加 Windows 10 trackpad fallback：统一 `trackpadScroll` 在 Windows 内部按普通 WHEEL/HWHEEL 注入，保持原生 1× delta；Win32 整数边界直接截断，不累计、不 `lround`，普通 `mouseScroll` 路径不变。插件 8 项 Dart 测试、父应用相关测试/analyze/macOS release build 已通过，麻烦复审。

windows/hardware_simulator_plugin.cpp
```diff
@@ -67,6 +67,22 @@ int logicalScrollToWindowsWheel(double logical_distance, bool invert) {
...
+    return 0;
+  }
+  const double direction = invert ? -1.0 : 1.0;
+  const double converted = logical_distance * direction;
+  const double clamped = (std::clamp)(
+      converted, -kMaxWindowsWheelDistance, kMaxWindowsWheelDistance);
+  // Win32 MOUSEINPUT::mouseData is integral. The trackpad path deliberately
+  // sends the canonical 1x delta directly at that API boundary: no mouse-wheel
+  // sensitivity multiplier, fractional accumulator, or nearest-value rounding.
+  return static_cast<int>(clamped);
```
hxyshadow: 💡 **截断会静默丢弃亚单位 delta —— 正是本 PR 在 macOS 侧修掉的那个 bug，在 macOS-client→Windows-host 方向又出现了。** 这里 `static_cast<int>(clamped)`（无 `lround`、无累加器）意味着原生 `dy=0.1` → `converted≈-0.1` → 截断成 `0` → `vertical_distance==0` → 不注入。而 macOS 的 `nativeTrackpadPointDelta` 对非零亚像素夹到 ±1 正是为了避免这种「0.1 被吃成 0」。

这是**有意的**设计（commit 说明明确「no accumulation, no nearest-value rounding」，避开了 ±1 夹紧带来的持续漂移——恰好呼应上轮 Claude 对 macOS point 轴 ±1 夹紧漂移的担忧），两种取舍各有代价：macOS 夹 ±1 = 不丢但会漂，Windows 截断 = 不漂但丢亚单位。Win32 `mouseData` 是整数，无法直接注入 0.1，所以只能二选一或引入残差累加器。

非阻塞:Windows 原生无精确捕获、消费端 cloudplayplus#442 按 macOS 门控,当前 macOS-client→Windows-host 精确滚动是边缘场景。但若将来 Windows host 精确滚动要面向用户,建议加个 fractional residual accumulator(跨帧累加不足 1 单位的余量),既不丢亚单位也不漂移。

hxyshadow (changes requested):
**🔴 REQUEST_CHANGES(需修改)**

## 🤖 Hermes reviewer — **APPROVE**（复审 · A2/Case-A 增量）

## Code Review: #41 — feat(macos): expose native trackpad scroll events

**Verdict: Approved ✅**（0 阻塞，1 条非阻塞 💡 · inline）
**Author:** @zhuhaichao518 | **Files:** 7 (+559 -2) | draft
**Reviewed against:** `5d56475`（上轮 APPROVE 针对 `33c9216`，作者随后推 1 个 commit `5d56475 fix(windows): replay trackpad scroll at native scale`，并 @我复审）

上轮两条阻塞（macOS point 轴 10× 过冲、符号约定不一致）+ HE-7 fallback inline 已在 `33c9216` 全部落地并 APPROVE。本轮增量**只有一个 commit**，纯补 Windows 10 **Host 端** trackpad fallback（`windows/hardware_simulator_plugin.cpp` +40，单文件）。作为迷你 PR 走了一遍 Step 4 + HE 系列（host 注入 shim）+ 双视角门（语义/跨文件 + 测试/生命周期）。

### ✅ 新 commit `5d56475` 复查

- **符号约定与鼠标滚轮路径自洽**：新增 `logicalTrackpadScrollToWindowsWheel(dx, false)` / `(dy, true)`，与兄弟函数 `logicalScrollToWindowsWheel`（`performMouseScroll` 用的那个，同样 `(dx,false)`/`(dy,true)`）方向完全一致——无轴间反向，注入方向与既有鼠标滚轮一致。✅
- **协议对称性**：新 method-channel case `trackpadScroll`（`:1503`）用 `std::get<double>(dx/dy)` + `result->Success(nullptr)`，与既有 `mouseScroll` 分支（`:1496`）对称。✅
- **边界健壮**：`std::isfinite` 守卫（NaN/Inf → 0）+ `std::clamp` 到 `±kMaxWindowsWheelDistance`（`WHEEL_DELTA*100`）防溢出；复用既有 `scroll()`/`hscroll()` → `send_input(MOUSEEVENTF_WHEEL/HWHEEL)` 路径，无新增权限面。✅
- **跨文件闭环**：Dart 侧 `performTrackpadScroll` 的 `on MissingPluginException { performMouseScroll }` fallback 现在只对无原生 case 的平台（Linux）触发；Windows 有了真实 native case，走精确 1× delta 注入，不再落 fallback。上轮 HE-7 的「fallback 名不副实」在 Windows 侧靠**真实实现**而非 catch 兜底解决。✅

### 💡 1 条非阻塞（inline · `windows/hardware_simulator_plugin.cpp:83`）

截断 `static_cast<int>(clamped)`（无 `lround`、无累加器）会**静默丢弃亚单位 delta**——原生 `dy=0.1` → 截断成 `0` → 不注入。这正是本 PR 在 macOS 侧（`nativeTrackpadPointDelta` 亚像素夹 ±1）修掉的「0.1 被吃成 0」，在 macOS-client→Windows-host 方向又出现了。**有意为之**（commit 明确「no accumulation, no nearest-value rounding」，避开了 ±1 夹紧的持续漂移，恰好呼应上轮 Claude 对 macOS point 轴漂移的担忧）：macOS 夹 ±1 = 不丢但漂，Windows 截断 = 不漂但丢亚单位。当前 Windows 无原生精确捕获、消费端按 macOS 门控，触发不到，非阻塞。若将来 Windows host 精确滚动面向用户，建议加 fractional residual accumulator（跨帧累加余量），两头都不吃亏。详见 inline。

### 💡 沿用（上轮已提，作者已显式 defer 到实测/决策，仍适用）

- **非消费式捕获的双源契约**（`macos ...swift:91` `return event`）：native monitor 不吞 AppKit 事件是对的，但同一次物理滚动 Flutter 仍派发自己的 `PointerPanZoom`。消费端 cloudplayplus#442 启用原生轨迹时须主动忽略 Flutter 自带 macOS 滚动源，否则双倍滚动——建议写进消费端 PR。
- 方向契约措辞张力（canonical「+y 向下」vs「跟随系统 scroll-direction 偏好」）、`unawaited(startTrackpadScrollCapture())` 丢弃返回值的启动失败反馈——作者已说明留待实测/决策，保持 ☑️。

### ✅ Looks Good

- Windows fallback 选型稳：不借用鼠标滚轮灵敏度乘子、直接在 Win32 整数边界送 1× delta，与 macOS 精确路径的「保真、不放大」原则一致。
- 单一 commit、单文件、纯加法，无回归，无 critical/warning 级残留。

上轮阻塞早已清除，本轮 Windows fallback 干净对称。维持 APPROVE。

---

## 🧠 Claude review — **REQUEST_CHANGES**

## 概述

PR 新增一套跨平台的「精密触控板滚动」通路：

- **Dart 层**：`TrackpadScrollEvent` / `TrackpadScrollPhase` / `TrackpadScrollCallback` 契约，第一个监听者启动原生捕获、最后一个移除后停止，以及 `performTrackpadScroll()` 注入 API（`MissingPluginException` 时回退到 `performMouseScroll`）。
- **macOS**：用 `NSEvent.addLocalMonitorForEvents(.scrollWheel)` 捕获（不吞事件），注入侧用 `CGEvent(scrollWheelEvent2Source:units:.pixel)` + `FixedPtDelta` 定点字段保留小数 + `IsContinuous`/`ScrollPhase`/`MomentumPhase`。
- **Windows**：无公开精密触控板注入 API，退化为普通高分辨率 WHEEL/HWHEEL，按 1× 原始量级。
- 测试：8 个 Dart method-channel 测试 + 1 个 macOS XCTest。

整体设计干净，方向约定（gesture → page direction）在两侧都有注释说明，回退路径和 no-op 语义处理得当。

## 阻塞问题

**1. Windows 侧 `logicalTrackpadScrollToWindowsWheel` 的截断会系统性丢量**（`windows/hardware_simulator_plugin.cpp:70-83`）

```cpp
return static_cast<int>(clamped);   // 向零截断
```

与同文件上方的 `logicalScrollToWindowsWheel`（用 `std::lround`）不一致。截断而非四舍五入，在连续手势中每帧平均损失 0.5px，方向恒为「偏小」——这与 PR 描述里「preserving the canonical 1x magnitude」的目标直接矛盾。macOS trackpad 常以 ~100Hz 出帧，累计漂移是可观的（数十 px/s 量级）。

此外 `|delta| < 1` 的帧会被完全丢弃（`performTrackpadScroll` 里 `!= 0` 才发），而 macOS 侧 `nativeTrackpadPointDelta` 对同样的输入是钳到 ±1 —— 两个平台对同一契约的行为不一致。

注释说这是刻意「不做累加、不做就近取整」，但保留 1× 量级和保留余数并不冲突：一个 per-axis 的余数累加器（把 `converted - trunc` 带到下一帧）既不引入鼠标滚轮灵敏度倍率，也不改变总量级，还能顺带解决亚像素帧被吞的问题。建议至少改成 `lround` + 余数进位。

## 建议（非阻塞）

**2. macOS 亚像素钳位可能过冲**（`HardwareSimulatorPlugin.swift`，`nativeTrackpadPointDelta`）
把 0.01 的 delta 提升为 point delta ±1 是合理妥协（`FixedPtDelta` 承载真值），但读 `PointDelta` 而不读定点字段的目标应用会看到 100× 放大。值得在函数注释里明确指出这个已知的取舍面向哪类消费者。

**3. Windows 完全忽略 `phase` / `isMomentum`**
方法通道传了这两个字段，Windows handler 直接丢弃。momentum 帧会被当作普通滚轮重放，可能与远端应用自己的惯性叠加。建议在 `HardwareSimulatorPlatform.performTrackpadScroll` 的文档注释里说明「phase/isMomentum 在部分平台上是 best-effort，可能被忽略」，避免调用方误以为跨平台保真。

**4. `addLocalMonitorForEvents` 只在本 app 获得焦点时触发**
对远程控制宿主场景大概率就是想要的行为，但这是与「native capture」字面含义不同的重要限制，建议写进 `addTrackpadScroll` 的 dartdoc。

**5. `startTrackpadScrollCapture` 的复用分支**（`swift`）
```swift
if trackpadScrollMonitor != nil { return flutterView != nil }
```
此分支里 `flutterView` 是 `weak`，若视图被释放会返回 `false` 但 monitor 仍挂着，且不会重新解析。虽然 `handleTrackpadScroll` 里有一次重解析逻辑，但这里直接返回 `false` 会让 Dart 侧误以为不支持。建议在这个分支里也尝试重新解析一次 view。

**6. `TrackpadScrollEvent` 标了 `@immutable` 但没有 `==`/`hashCode`**
纯粹是一致性问题；如果调用方要做去重/diff 会不方便。

**7. Windows handler 缺参数校验**
`args->find(...)->second` 未检查 `end()`，`std::get<double>` 在类型不符时抛 `bad_variant_access`。与既有 `mouseScroll` 分支风格一致，所以不算回归，但这条新路径同样值得顺手加上防护。

## 测试覆盖

- Dart 侧覆盖不错：小数保真、`MissingPluginException` 回退、监听者生命周期（含重复添加去重）、原生事件解码。
- macOS 只覆盖了 `nativeTrackpadPointDelta`；`trackpadScrollPhaseName` / `nativeTrackpadScrollPhase` / `nativeTrackpadMomentumPhase` 三个纯映射函数很容易加测试，尤其是 `NSEvent.Phase` 是 OptionSet、优先级顺序（cancelled > ended > began > …）是有语义的，值得锁住。
- Windows 的转换函数（本次最可能出问题的地方）完全没有测试。

## 安全性 / 性能

- 无新增权限面；macOS 注入走 `cghidEventTap`，与既有 `performMouseScroll` 同级，需要既有的 Accessibility 授权，无提权。
- local monitor 回调里不 consume 事件（`return event`），不会破坏 AppKit 正常滚动。
- 每帧一次 `invokeMethod`，在高频 momentum 阶段（~100Hz）有一定 channel 开销，但 payload 很小，可接受。
- `deinit` 里调用 `NSEvent.removeMonitor` 需在主线程；plugin 通常主线程释放，风险低。

zhuhaichao518:
@hxyshadow 已在 `abcd29c` 处理最新复审的 Windows 截断丢量问题：Win10 trackpad fallback 现在按 `WHEEL_DELTA / 100 = 1.2` 做网页像素到 Win32 wheel unit 的换算，X/Y 两轴分别跨帧保留小数余量，只在最终边界向零截断，不使用 `lround`。连续五帧 `1px` 输出 `1,1,1,1,2`，累计网页距离保持 `5px`；亚单位帧不再永久丢失。新增 Windows gtest 覆盖累计、慢速帧和垂直方向。插件 8 项 Dart 测试及父应用 33 项定向测试、analyze、macOS release build 均通过，麻烦基于最新提交复审。

hxyshadow (approved):
**✅ APPROVE(通过)**

## Code Review: #41 — feat(macos): expose native trackpad scroll events（复审 · Case-A 增量）

**Verdict: Approved ✅**（0 阻塞）
**Author:** @zhuhaichao518 | **Files:** 10 (+634 -2) | draft
**Reviewed against:** `abcd29cda9aec8deac2fbf141fee70af52adb475`（上轮 REQUEST_CHANGES 针对 `5d56475`，作者随后推 1 个 commit `abcd29c fix(windows): preserve trackpad wheel remainder`）

本轮增量只有一个 commit（4 文件 +95/-20），精确落地了我上轮那条 Windows 截断丢量的 💡——逐条在 `abcd29c` 核过：

### ✅ 上轮 finding 已修复

- **Windows 截断静默丢弃亚单位 delta**（上轮 inline，`performTrackpadScroll`）：旧的 `logicalTrackpadScrollToWindowsWheel`「`static_cast<int>(clamped)` 直接截断、无累加器、无 lround」已整体删除（repo 内 0 残留调用者），换成新的 `TrackpadScrollAccumulator`（`windows/trackpad_scroll_accumulator.h`）。做法正是我建议的**逐轴跨帧 fractional residual accumulator**：
  - 新增像素→Win32 wheel-unit 换算常量 `kWindowsTrackpadWheelUnitsPerPixel = WHEEL_DELTA/100 = 1.2`，累计页面距离保持 1:1；
  - X/Y 两个 `thread_local` 累加器各自保留小数余量，只在最终边界向零截断，`residual_ = clamped - whole_units` 带到下一帧；
  - `1e-9` 边界容差专门防 `5 * 1.2` 这类精确有理数因二进制浮点表示落在整数正下方——是对的，只影响进位判定不影响量级。
  - 我把两个新 gtest 的断言用独立 Python 模型逐帧重算全部命中：`PreservesCumulativeBrowserPixelDistance` 连续五帧 `1px` → `1,1,1,1,2`（累计 6 = 5×1.2，零丢失）；`CarriesSubunitFramesPerAxis` 八帧 `0.1px` 全 `0`、第九帧进位 `1`（亚单位帧不再永久丢失），纵轴取反语义 `-1,-1,-1,-1,-2` 也对。macOS 夹 ±1（不丢但漂）vs Windows 截断丢量 的**两平台契约不一致**现已消除——两头都用「保余量、不改量级」的语义。

### 💡 沿用（非阻塞，作者已显式 defer 到实测/决策）

- **非消费式捕获的双源契约**（`swift:91` `return event`）：native monitor 不吞 AppKit 事件是对的，但同一次物理滚动 Flutter 仍会派发自己的 `PointerPanZoom`；消费端 cloudplayplus#442 启用原生轨迹时需主动忽略 Flutter 自带 macOS 滚动源，否则双倍滚动。属消费端 PR 范畴。
- **`startTrackpadScrollCapture` 结果被丢弃**（`unawaited`，view 为 nil 时静默失败）：作者已说明属行为设计，保持现状等实测。

### 微小观察（不阻塞、无需处理）

- `TrackpadScrollAccumulator` 是 `thread_local`，residual 跨手势边界持续保留。相邻两次反向手势时，上一手势遗留的 <1 wheel-unit 余量会带入下一手势首帧——量级 <1px，滚动场景不可感知且自纠正，无需清零。

修复干净、对称、有 gtest 回归兜底，正好把上轮我提的两平台不对称补齐。draft 状态，方向与实现都稳，维持 APPROVE。

---

## PR #41 — `feat(macos): expose native trackpad scroll events`

### 概览

新增一条跨平台「精密触控板滚动」通路：

- **捕获侧**：Dart 层 `addTrackpadScroll`/`removeTrackpadScroll` + 引用计数（首个监听者启动、最后一个移除后停止原生捕获）；macOS 通过 `NSEvent.addLocalMonitorForEvents(.scrollWheel)` 捕获精密滚动，转换为 Flutter view 坐标 + page-direction delta + phase/momentum，回调 `onTrackpadScroll`。
- **注入侧**：`performTrackpadScroll(dx, dy, phase, isMomentum)`；macOS 用 `CGEvent(scrollWheelEvent2Source:)` 重建连续事件并写入 16.16 定点 delta 与 scroll/momentum phase；Windows 10 无公开精密触控板注入 API，退化为普通 WHEEL/HWHEEL，并用 `TrackpadScrollAccumulator` 按轴累积小数余量。
- 未实现的平台安全降级：`startTrackpadScrollCapture` 返回 false，`performTrackpadScroll` 捕获 `MissingPluginException` 回落到 `performMouseScroll`。

### 正确性

我verified了几个容易出问题的点，结论是没有阻塞性缺陷：

- 方向约定一致：Windows 侧 `dx` 用 `invert=false`、`dy` 用 `invert=true`，与既有 `performMouseScroll` 完全一致；macOS 侧捕获取负（AppKit gesture → page direction）、注入再取负还原，闭环自洽。
- Windows 的 `startTrackpadScrollCapture` / `stopTrackpadScrollCapture` 会落到 `hardware_simulator_plugin.cpp:1977` 的 `result->NotImplemented()`，Dart 端的 `MissingPluginException` catch 能正确兜住，不会抛到调用方。
- `TrackpadScrollAccumulator::Convert` 的 clamp 发生在加上 residual 之后，且 `residual_ = clamped - whole_units`，饱和时残差不会失控累积；`|residual| < 1` 恒成立。`1e-9` 的边界容差配合 `copysign` 处理了 `5 * 1.2` 这类二进制表示落在整数下方的情况，方向对称。
- `@immutable` 有对应的 `package:flutter/foundation.dart` import；RunnerTests 用的是 `@testable import hardware_simulator`，所以 `static func nativeTrackpadPointDelta` 保持 internal 可见性是可以的。
- `wheelCount: horizontalPoint != 0 ? 2 : 1` 安全：`nativeTrackpadPointDelta` 对任何非零有限输入都不会返回 0。

### 建议（非阻塞）

1. **macOS 缺少与 Windows 对称的余量累积**（`HardwareSimulatorPlugin.swift`，`nativeTrackpadPointDelta`）。对连续的 0.4px 帧，每帧的 `PointDelta` 都被抬到 ±1，对**只读 point/wheel 字段**的接收端会产生约 2.5× 的距离放大。定点字段是准确的，主流消费者（浏览器、AppKit 滚动视图）读定点，所以影响有限；但把 Windows 那个 accumulator 的思路移植过来（每轴累积 point delta 余量）会让两端行为一致。注释里已经承认了这个取舍，至少值得留个 TODO。

2. **Dart 侧 start/stop 的竞态**（`hardware_simulator_method_channel.dart`）。两处都是 `unawaited(...)`，快速 add→remove→add 序列下，`stopTrackpadScrollCapture` 可能在后一次 `startTrackpadScrollCapture` 之后才到达原生层，导致捕获被意外关闭。用一个串行化的 Future 链或代际计数器（generation counter）即可消除。实际触发概率低，但监听者由上层 widget 生命周期驱动时并非不可能。

3. **Windows 忽略 `phase` / `isMomentum`**。方法通道会传这两个字段，C++ 端直接丢弃。功能上可接受（legacy wheel 无处安放），但由此带来的副作用是 `residual_` 会跨手势无限期保留 —— 建议在收到 `began` 时把两个 accumulator 的残差清零，避免上一手势的尾巴影响下一手势的第一帧。同时 `thread_local` 静态状态放在自由函数里，比挂在 plugin 实例上更难测试和重置。

4. **Windows 参数解析未做防御**：`args->find(...)` 的返回值未与 `end()` 比较即解引用，`std::get<double>` 也可能抛 `bad_variant_access`。这与既有的 `mouseScroll` 分支写法一致，所以算是遵循项目现状而非本 PR 引入的退步 —— 只是新增代码复制了同一个模式。

5. **macOS 默认 phase 语义**：Dart 侧 `phase` 默认 `none`，此时注入的事件 scrollPhase 与 momentumPhase 均为 0。对于「只想推一段像素、不关心手势生命周期」的调用方，这是合理的；不过文档注释里说明一下「不带 phase 时部分应用可能不触发惯性/橡皮筋效果」会更好。

6. 未见 `CHANGELOG.md` / `pubspec.yaml` 版本更新 —— 若该插件对外发布，属于公开 API 新增，通常需要一起 bump。

### 测试覆盖

覆盖得相当扎实：Dart 侧覆盖了小数保真、MissingPluginException 回落、首/末监听者的捕获生命周期、以及原生事件解码；C++ 侧覆盖了累加器的累积距离与逐轴余量携带；Swift 侧覆盖了 point delta 的符号、亚像素抬升与 NaN。缺口主要是 macOS `handleTrackpadScroll` 的坐标翻转（`isFlipped` 分支）与 `trackpadScrollPhaseName` 的位组合优先级没有单测，这两处都是纯函数逻辑，容易补。

### 性能与安全

- 每帧新建一次 `CGEventSource(stateID: .hidSystemState)`，在 120Hz 惯性流下会有些许开销，可以提升为成员变量缓存。
- macOS 用的是 **local** monitor 且返回原 event，不吞事件、不需要辅助功能授权即可捕获，作用域限于本应用窗口 —— 这是正确且克制的选择。注入侧仍走 `cghidEventTap`，权限要求与既有鼠标注入一致，未扩大攻击面。

### 结论

设计清晰、平台差异处理得当、注释解释了每个非显然的取舍，测试覆盖到位。上述几点都是打磨性质的改进，不影响合并。

zhuhaichao518:
@hxyshadow 最新 `162ed87` 按实测策略调整职责：macOS precise→Windows 的 ×2 放在主仓 Dart 归一化层；Windows native adapter 已移除 `WHEEL_DELTA / 100 = 1.2`，现在仅做方向、范围限制、向零截断和双轴小数余量累计，不使用 `lround`。插件 8 项测试、C++ 语法检查、父应用 78 项相关测试/analyze/macOS release build 均通过，麻烦复审。

hxyshadow (commented):
**💬 COMMENT(评论)**

## Code Review: #41 — feat(macos): expose native trackpad scroll events（复审 · A2/Case-A 增量）

**Verdict: Approved ✅**（0 阻塞）
**Author:** @zhuhaichao518 | **Files:** 10 (+625 -3) | draft
**Reviewed against:** `162ed873df0ca5cfc37efdbb9ef150667fea9f46`（上轮 APPROVE 针对 `abcd29c`，作者随后推 1 个 commit `162ed87 fix(windows): keep trackpad scaling in Dart`，并 @我复审）

本轮增量只有一个 commit，改 3 个 Windows 文件（`hardware_simulator_plugin.cpp` / `trackpad_scroll_accumulator.h` / `test/hardware_simulator_plugin_test.cpp`），把 `WHEEL_DELTA / 100 = 1.2` 的「网页像素 → Win32 wheel unit」换算从 Windows native adapter 移到主仓 Dart 归一化层。`TrackpadScrollAccumulator` 现在退化为纯 1:1 的跨帧小数余量搬运器。

### ✅ 本轮增量核过

- **职责边界更干净。** native adapter 不再持有 `output_units_per_input_unit_` 缩放因子，构造函数从 `(units_per_pixel, max)` 收窄为 `explicit (max)`。`Convert` 里 `logical_distance * direction * units_per_pixel + residual_` 简化为 `logical_distance * direction + residual_`，余量搬运逻辑（clamp → tolerance → truncate → residual carry）完全保留，未被这次简化破坏。macOS precise→Windows 的 ×2 放主仓 Dart 后，插件侧就是纯粹的「1× 输入 → 整数边界 + 亚单位跨帧累计」，符合「cross-platform magnitude scaling belongs to Dart」的新注释。
- **跨文件影响干净（git grep @162ed87 全核过）。** 删掉的 `kWindowsTrackpadWheelUnitsPerPixel` 常量与 `output_units_per_input_unit_` 成员均 0 残留引用；两处 `thread_local` accumulator 构造点、以及两个 gtest 构造点全部同步改成单参构造；`kMaxWindowsWheelDistance` 正确保留——它同时被普通 `logicalScrollToWindowsWheel`（mouseScroll 路径）和 accumulator 的 clamp 复用，删不得。普通 `performMouseScroll` 路径未受影响。
- **测试断言随缩放因子移除同步修正，逐条核过全对。**
  - `PreservesIntegralInputDistance`（原 `PreservesCumulativeBrowserPixelDistance`）：1:1 下 `Convert(1)×5` → `1,1,1,1,1`（旧 ×1.2 是 `1,1,1,1,2`），改对。
  - `CarriesSubunitFramesPerAxis`：`Convert(0.1)` 在 1:1 下每帧攒 0.1，循环从 `i<8` 调成 `i<9`（0.9 < 1 不进位），第 10 帧 `1.0` → `1`；纵轴 `Convert(1,true)×4` → `-1,-1,-1,-1`（1:1 无余量堆积，旧 ×1.2 第 5 帧会到 `-2`），断言由 `-2` 改回 `-1`，改对。亚像素不丢、方向取反、非有限值守卫这三点仍被锁住。

### 💡 沿用（上轮已提，作者已显式 defer 到实测/决策，非阻塞）

- **`residual_` 跨手势不清零**（Claude 上轮 #3）：Windows 侧丢弃 `phase`/`isMomentum`，accumulator 收不到 `began` 信号，上一手势尾巴的亚单位余量会带进下一手势第一帧。缩放因子移除后 accumulator 成了唯一的量级搬运点，这条反而更集中——将来若 Windows host 精确滚动面向用户，建议在收到手势起始时 reset 双轴 residual。当前 Windows 无原生精确捕获、消费端按 macOS 门控，触发不到。
- **Dart 侧 start/stop 竞态**（两处 `unawaited`）、**非消费式捕获的双源契约**（`swift:91` `return event`，Magic Mouse 下 Flutter 自身 `PointerScrollEvent` 与 native monitor 双源）：均属行为设计，作者保持现状等实测/决策。

draft 状态，本轮是一次干净的职责重构：缩放上移、accumulator 收窄为纯余量搬运、测试同步修正且逐条正确。维持 APPROVE。

---

## Overview

Adds a cross-platform precision-trackpad path to the plugin:
- **Capture**: new `TrackpadScrollEvent` / `TrackpadScrollPhase` contract, `addTrackpadScroll`/`removeTrackpadScroll` with refcounted native start/stop; macOS implements it via a local `NSEvent.scrollWheel` monitor (non-consuming).
- **Injection**: `performTrackpadScroll` → macOS rebuilds a continuous `CGEvent` with fixed-point deltas + scroll/momentum phase; Windows replays through legacy `WHEEL`/`HWHEEL` with a new per-axis fractional accumulator.
- Fallback to `performMouseScroll` on `MissingPluginException`; unsupported platforms are no-ops.

The design is sound and the layering (magnitude scaling in Dart, direction/phase normalization in the platform layer) is the right call. Dart-side test coverage is genuinely good.

## Main concerns

**1. The `dx`/`dy` unit contract differs per platform, but the docs claim it's unified.**
`performTrackpadScroll`'s doc says only "positive x scrolls right, positive y scrolls down". On macOS the value is fed straight into `scrollWheelEventFixedPtDeltaAxis*` — i.e. **pixels**. On Windows it goes to `hscroll`/`scroll` clamped against `WHEEL_DELTA * 100` and the accumulator comment calls it "fractional **wheel-unit** remainder" — i.e. **1 unit ≈ 1/120 notch**. The same `dy = 100` means 100 px on macOS and ~0.83 notches on Windows. Since the whole point of moving scaling into Dart is that callers do it, the unit must be stated in the public dartdoc (`lib/hardware_simulator.dart:40-44` and the platform interface). Right now every consumer has to read the native source to know what to pass.

**2. macOS amplifies sub-pixel deltas ~10× on the integer axes; Windows correctly accumulates.**
`nativeTrackpadPointDelta` (`macos/Classes/HardwareSimulatorPlugin.swift`) turns any non-zero sub-pixel delta into ±1:
```swift
var rounded = (-logicalDelta).rounded()
if rounded == 0 { rounded = logicalDelta.isLess(than: 0) ? 1 : -1 }
```
A momentum tail streaming `0.05` px per frame at 120 Hz emits `pointDelta = 1` every frame — 20× the intended distance. It also rounds `1.4 → 1` and `1.6 → 2` with no residual carry, so `pointDelta` and `fixedPtDelta` diverge cumulatively. Clients that read `scrollWheelEventPointDeltaAxis*` (Chromium/Electron do, on several code paths) will see drift rather than the true motion.

Windows already solved exactly this problem with `TrackpadScrollAccumulator`. The macOS injector should carry a residual the same way (per-axis, reset on `began`), so the integer axes track the fixed-point value instead of ratcheting.

**3. Windows silently discards `phase` and `isMomentum`, and the accumulator residual never resets.**
`performTrackpadScroll(double dx, double dy)` doesn't even receive them. Two consequences:
- No gesture boundary → the `thread_local` residual persists across gestures. A leftover `0.99` from a finished gesture can emit a spurious unit on the first frame of the next one. Passing `phase` through and clearing `residual_` on `began`/`ended` fixes it cheaply (needs a small `Reset()` on the accumulator).
- The `thread_local` accumulators are function-local statics tied to whichever thread the platform channel runs on. That's fine today, but a brief comment on why thread-local (rather than instance state) would help.

## Smaller notes

- `startTrackpadScrollCapture()`'s return value is ignored by `addTrackpadScroll` (`lib/hardware_simulator_method_channel.dart:118-120`). On macOS it returns `false` when `registrar.view` isn't attached yet — a listener registered early gets silently no events with no retry. Either surface the result (make `addTrackpadScroll` return `Future<bool>`) or retry/lazy-resolve the view.
- `_asDouble` maps non-numeric payloads to `0` silently; combined with the `arguments is Map` guard, a malformed native frame becomes a plausible-looking zero-delta event. Not exploitable, just hard to debug — a `assert`/log in debug would help.
- `nativeTrackpadPointDelta` is `internal`; the new `RunnerTests` case will only compile with `@testable import`. Worth confirming that target already uses it.
- The Windows gtest additions look correct (`0.1 × 10 → 1` exercises the `1e-9` boundary tolerance exactly as intended), but the PR's validation list only mentions a "header syntax check" — the Windows test binary apparently wasn't run. Please confirm it builds and passes before merge, since `windows/test/` includes the header by bare name.
- Stray unrelated hunk: the removed blank line at `windows/hardware_simulator_plugin.cpp:56`. Trivial, but it's noise in the diff.
- `max_output_distance_` can be `const double`.

## Security / performance

Nothing security-relevant. Perf is fine: the macOS monitor returns the event unmodified (doesn't swallow AppKit input) and early-outs on `hasPreciseScrollingDeltas` and zero phase, so ordinary mouse wheels cost almost nothing. Capture is only installed while a listener exists.

## Verdict rationale

Items 1–3 are quality/correctness issues rather than crashes, and #2 is a deliberate, commented, tested tradeoff — so I'm not hard-blocking. But I'd want at least the unit documentation (#1) and the macOS residual carry (#2) addressed before this ships, since #2 will produce visible scroll drift on real macOS clients.

zhuhaichao518:
@hxyshadow 最新提交 `162ed87` 已确认 Windows trackpad native 不做任何倍率：只做方向、安全范围、逐轴小数余量累计和 Win32 整数边界注入；倍率全部由 Viewer 发送端决定。请对当前 head 补一个 GitHub 正式 APPROVE，收到后我会立即 squash merge，并将最终 main commit 回填 cloudplayplus#442。

zhuhaichao518:
@hxyshadow 再 ping 一下：当前 head 仍是 `162ed87`，代码和验证结果都没有变化；你上一轮 review 内容已是 **Approved / 0 阻塞**，但 GitHub review state 是 `COMMENTED`，所以分支还没有正式 approval。麻烦直接对当前 head 提交一次 GitHub **APPROVE**，我会随后按已批准流程立即 squash merge。

PR #42, opened 2026-08-03
fix(macos): refresh cursor textures and hotspot

zhuhaichao518:
## Summary

- detect cursor texture transitions on mouse button down/up with bounded follow-up probes
- serialize cursor bitmaps consistently and convert hotspots using the bitmap representation scale
- remove invalid cursor nil checks and avoid sending an optional default cursor index
- add macOS tests for event coverage and hotspot conversion

## Verification

- user-verified in CloudPlayPlus with Slay the Spire 2 and the brush cursor
- `flutter test` (8/8 passed)
- macOS XCTest (5/5 passed)
- `./build.sh --macos --release`

Release artifact: `cloudplayplus/build/macos/Build/Products/Release/cloudplayplus.app`

### DIFF
--- a/macos/Classes/HardwareSimulatorPlugin.swift
+++ b/macos/Classes/HardwareSimulatorPlugin.swift
@@ -2154,7 +2154,18 @@ public class HardwareSimulatorPlugin: NSObject, FlutterPlugin {
   var jsHashWithImageHash = Dictionary<String, UInt32>()
   var cursorHashesByCallback = Dictionary<Int, Set<UInt32>>()
   var hookAllCursorImage = Dictionary<Int, Bool>()
-  let cursorMonitorMask: NSEvent.EventTypeMask = [.mouseMoved, .leftMouseDragged, .rightMouseDragged, .otherMouseDragged]
+  let cursorMonitorMask: NSEvent.EventTypeMask = [
+    .mouseMoved,
+    .leftMouseDragged,
+    .rightMouseDragged,
+    .otherMouseDragged,
+    .leftMouseDown,
+    .leftMouseUp,
+    .rightMouseDown,
+    .rightMouseUp,
+    .otherMouseDown,
+    .otherMouseUp,
+  ]
 
   func JSHash(buffer: [UInt8], size: Int) -> UInt32 {
     var hash: UInt32 = 1315423911
@@ -2162,6 +2173,91 @@ public class HardwareSimulatorPlugin: NSObject, FlutterPlugin {
         hash ^= ((hash << 5) &+ UInt32(buffer[i]) &+ (hash >> 2))
     }
     return hash & 0x7FFFFFFF
+  }
+
+  static func cursorHotSpotInPixels(
+    _ hotSpot: NSPoint,
+    pixelWidth: Int,
+    pixelHeight: Int,
+    pointSize: NSSize
+  ) -> (x: UInt32, y: UInt32) {
+    let scaleX = pointSize.width > 0
+      ? CGFloat(pixelWidth) / pointSize.width
+      : 1
+    let scaleY = pointSize.height > 0
+      ? CGFloat(pixelHeight) / pointSize.height
+      : 1
+
+    return (
+      x: cursorCoordinate(hotSpot.x * scaleX),
+      y: cursorCoordinate(hotSpot.y * scaleY)
+    )
+  }
+
+  private static func cursorCoordinate(_ value: CGFloat) -> UInt32 {
+    guard value.isFinite, value > 0 else { return 0 }
+    return UInt32(min(value.rounded(), CGFloat(UInt32.max)))
+  }
+
+  private func appendUInt32BE(_ value: UInt32, to bytes: inout [UInt8]) {
+    bytes.append(UInt8((value >> 24) & 0xFF))
+    bytes.append(UInt8((value >> 16) & 0xFF))
+    bytes.append(UInt8((value >> 8) & 0xFF))
+    bytes.append(UInt8(value & 0xFF))
+  }
+
+  private func encodeCursorBitmap(
+    image: NSImage,
+    hotSpot: NSPoint
+  ) -> (hash: UInt32, payload: Data)? {
+    guard let bitmapRep = image.representations.first(where: {
+      $0 is NSBitmapImageRep
+    }) as? NSBitmapImageRep,
+      let pixels = getBitMapInt8(bitmapRep: bitmapRep) else {
+      return nil
+    }
+
+    let messageHash = JSHash(buffer: pixels, size: pixels.count)
+    let pointSize = bitmapRep.size.width > 0 && bitmapRep.size.height > 0
+      ? bitmapRep.size
+      : image.size
+    let hotSpotPixels = Self.cursorHotSpotInPixels(
+      hotSpot,
+      pixelWidth: bitmapRep.pixelsWide,
+      pixelHeight: bitmapRep.pixelsHigh,
+      pointSize: pointSize
+    )
+
+    var bytes: [UInt8] = [9]
+    appendUInt32BE(UInt32(clamping: bitmapRep.pixelsWide), to: &bytes)
+    appendUInt32BE(UInt32(clamping: bitmapRep.pixelsHigh), to: &bytes)
+    appendUInt32BE(hotSpotPixels.x, to: &bytes)
+    appendUInt32BE(hotSpotPixels.y, to: &bytes)
+    appendUInt32BE(messageHash, to: &bytes)
+    bytes.append(contentsOf: pixels)
+    return (messageHash, Data(bytes))
+  }
+
+  private func handleCursorMonitorEvent(_ event: NSEvent) {
+    checkMouseCursor()
+    sendCursorPositionToAll()
+
+    switch event.type {
+    case .leftMouseDown, .leftMouseUp,
+         .rightMouseDown, .rightMouseUp,
+         .otherMouseDown, .otherMouseUp:
+      // Games commonly swap their cursor after handling the button event or
+      // on the following render frame. Bounded follow-up probes catch that
+      // transition without introducing a permanent high-frequency timer.
+      for delay in [16, 50] {
+        DispatchQueue.main.asyncAfter(deadline: .now() + .milliseconds(delay)) {
+          [weak self] in
+          self?.checkMouseCursor()
+        }
+      }
+    default:
+      break
+    }
   }
 
   func getBitMapInt8(bitmapRep: NSBitmapImageRep) -> [UInt8]?{
@@ -2280,16 +2376,15 @@ public class HardwareSimulatorPlugin: NSObject, FlutterPlugin {
   }
 
   private func checkMouseCursor() {
-   
-    let currentCursor = NSCursor.currentSystem 
-    if(currentCursor == nil){
-        return;
+    guard !cursorChangedCallbacks.isEmpty,
+      let currentCursor = NSCursor.currentSystem else {
+      return
     }
-    
-    let cursorImage = currentCursor?.image
-    let hotSpot = currentCursor?.hotSpot
 
-    let cursorImageHashes = sha256ForAllBitmapReps(in: cursorImage!)
+    let cursorImage = currentCursor.image
+    let hotSpot = currentCursor.hotSpot
+
+    let cursorImageHashes = sha256ForAllBitmapReps(in: cursorImage)
 
     if(cursorImageHashes == previousCursorImageHashes){
         return;
@@ -2298,8 +2393,7 @@ public class HardwareSimulatorPlugin: NSObject, FlutterPlugin {
     previousCursorImage = cursorImage
 
     //system default
-    if defaultCursorHasher?.getHashMap().keys.contains(cursorImageHashes) ?? false {
-      let cursorIndex = defaultCursorHasher?.getHashMap()[cursorImageHashes]
+    if let cursorIndex = defaultCursorHasher?.getHashMap()[cursorImageHashes] {
         //print("default: \(cursorIndex!)");
         for callbackID in cursorChangedCallbacks {
             if !(hookAllCursorImage[callbackID] ?? false) {
@@ -2312,26 +2406,20 @@ public class HardwareSimulatorPlugin: NSObject, FlutterPlugin {
                 methodChannel?.invokeMethod("onCursorImageMessage", arguments: message)
             } else {
                 // For hookAll=true, treat it as if it's not a default cursor
-                let imagedataInt8 = getBitMapInt8(bitmapRep: cursorImage!.representations[0] as! NSBitmapImageRep)
-                let messageHash = JSHash(buffer: imagedataInt8!, size: imagedataInt8!.count)
+                guard let encoded = encodeCursorBitmap(
+                  image: cursorImage,
+                  hotSpot: hotSpot
+                ) else {
+                  continue
+                }
+                let messageHash = encoded.hash
                 jsHashWithImageHash[cursorImageHashes] = messageHash
-
-                var int8Image:[UInt8] = [9];
-                int8Image.append(contentsOf:[0,0,0,UInt8(cursorImage!.representations[0].pixelsWide)])
-                int8Image.append(contentsOf:[0,0,0,UInt8(cursorImage!.representations[0].pixelsHigh)])
-                int8Image.append(contentsOf:[0,0,0,UInt8(hotSpot!.x)])
-                int8Image.append(contentsOf:[0,0,0,UInt8(hotSpot!.y)])
-                int8Image.append(UInt8((messageHash >> 24) & 0xFF))
-                int8Image.append(UInt8((messageHash >> 16) & 0xFF)) 
-                int8Image.append(UInt8((messageHash >> 8) & 0xFF))
-                int8Image.append(UInt8(messageHash & 0xFF))
-                int8Image.append(contentsOf: imagedataInt8!)
 
                 let message: [String: Any] = [
                     "callbackID": callbackID,
                     "message": CursorConstants.cursorUpdatedImage,
                     "msg_info": messageHash,
-                    "cursorImage": FlutterStandardTypedData.init(bytes: Data(int8Image))
+                    "cursorImage": FlutterStandardTypedData.init(bytes: encoded.payload)
                 ]
                 methodChannel?.invokeMethod("onCursorImageMessage", arguments: message)
                 markCursorHashSeen(callbackID, messageHash)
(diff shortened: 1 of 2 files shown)


zhuhaichao518:
@hxyshadow review

macos/Classes/HardwareSimulatorPlugin.swift
```diff
@@ -2164,6 +2175,91 @@ public class HardwareSimulatorPlugin: NSObject, FlutterPlugin {
...
+    sendCursorPositionToAll()
+
+    switch event.type {
+    case .leftMouseDown, .leftMouseUp,
+         .rightMouseDown, .rightMouseUp,
+         .otherMouseDown, .otherMouseUp:
+      // Games commonly swap their cursor after handling the button event or
+      // on the following render frame. Bounded follow-up probes catch that
+      // transition without introducing a permanent high-frequency timer.
+      for delay in [16, 50] {
```
hxyshadow: 💡 **Suggestion:** The bounded probe schedule `[16, 50]` ms is a nice tradeoff vs. a permanent high-frequency timer, but it's an inline magic array. Consider hoisting it to a named `static let` (e.g. `cursorTransitionProbeDelaysMs = [16, 50]`) so the coverage/latency tradeoff is discoverable and tunable in one place. Note the inherent limit worth documenting: a game that swaps its cursor >50 ms after the button event (e.g. a render hitch, or ~4+ frames at 60 fps) won't be caught until the next mouse-move — acceptable for now, just flagging the boundary.

hxyshadow (approved):
**✅ APPROVE(通过)**

## Code Review: #42 — fix(macos): refresh cursor textures and hotspot

**Verdict: Approved ✅** (0 issues, 1 suggestion)
**Author:** @zhuhaichao518 | **Files:** 2 (+173 -87)
**Reviewed against:** `a00f64e` (HEAD)

First-pass review — no prior reviewer, verified on disk against HEAD with full cross-file impact analysis. This is a solid correctness + robustness refactor. Every change either fixes a real crash path or improves the wire encoding, and the new XCTests pin the two trickiest units.

### ✅ Looks Good
- **Crash-path hardening.** `encodeCursorBitmap` replaces the old `representations[0] as!` force-cast with a safe `.first(where: { $0 is NSBitmapImageRep }) as? NSBitmapImageRep` + `guard`. Same `guard`-and-`return nil` shape is now applied at all four call sites (default-hookAll branch, cache branch, fresh-image branch, and the `hookCursorImage` initial push), so a rep-less cursor degrades gracefully instead of trapping.
- **Hotspot scaling is now principled and NaN-safe.** `cursorHotSpotInPixels` derives the scale from the bitmap representation (`pixelsWide / pointSize.width`) instead of the old `NSScreen.backingScaleFactor` + `<=32 ? 1.0` heuristic, and `cursorCoordinate` clamps non-finite / negative to 0 and caps at `UInt32.max` via `min(...rounded(), UInt32.max)`. The two XCTests (`…UsesBitmapRepresentationScale` asserting 7.5·(64/32)=15, and `…ClampsInvalidCoordinates` asserting NaN/negative→0) lock both paths.
- **Wire header widened to real BE UInt32.** `appendUInt32BE(UInt32(clamping: pixelsWide))` replaces the old `[0,0,0,UInt8(pixelsWide)]` single-byte-with-3-zero-pad, which silently truncated any dimension/hotspot ≥256. Layout is byte-identical for the common ≤255 px cursor, and the Dart consumer (`hardware_simulator_method_channel.dart:54`) forwards `cursorImage` opaquely, so no consumer change is needed. Nice that hotspot is now folded into the shared `encodeCursorBitmap` too, killing three duplicated hand-rolled serializers.
- **Dead-code removal is genuinely dead.** The removed `if(cursor == nil || cursor.image == nil)` in `CursorHasher.calculateHashesForDefaultCursors` operated on a `[NSCursor: Int]` literal with non-optional keys — the check could never fire. Similarly the `defaultCursorHasher?…keys.contains` → `if let cursorIndex = …[hash]` rewrite is a clean optional-binding simplification with identical semantics.
- **Button-transition probing is well-bounded.** Adding the 6 mouse down/up masks + the `[16, 50] ms` follow-up `checkMouseCursor()` probes catches the common "game swaps cursor on the frame after the click" case without a permanent high-frequency timer. Both probes use `[weak self]` (no retain cycle) and `checkMouseCursor`'s early-out guards make the extra calls idempotent/cheap. The `cursorMonitorMask` change is covered by the new `testCursorMonitorIncludesButtonTransitions`, and all three monitor-registration sites (`hookCursorImage`, the two `hookCursorPosition` paths) reuse the same mask constant, so coverage is symmetric.

### 💡 Suggestions
- **HardwareSimulatorPlugin.swift:2252** — hoist the `[16, 50]` probe schedule to a named `static let` and document the >50 ms boundary (see inline). Non-blocking.

Cross-platform note: the button-transition probe is specific to the macOS global-monitor model; Windows drives cursor updates through `SetCursor`/`onCursorImageMessage` in `windows/hardware_simulator_plugin.cpp` and needs no equivalent — no sibling-platform gap introduced.

---

## Overview

Fixes macOS cursor capture in `HardwareSimulatorPlugin.swift`:
1. Adds mouse **down/up** to the global event monitor mask + two bounded follow-up probes (16 ms / 50 ms) so game cursor swaps that happen after a click are detected.
2. Extracts the duplicated cursor-bitmap serialization into `encodeCursorBitmap(image:hotSpot:)` (4 call sites), and derives hotspot scale from the bitmap rep (`pixelsWide / size.width`) instead of `NSScreen.backingScaleFactor` + the `<=32` heuristic.
3. Removes force-unwraps and dead nil checks; adds 3 XCTests.

Net −87/+173 with a large dedup — good direction.

## Strengths

- **Real bug fixed**: the old `UInt8(pixelsWide)` / `UInt8(hotSpot.x)` truncated silently for cursors >255 px or hotspots >255. `appendUInt32BE` + `UInt32(clamping:)` is correct and wire-compatible (the old code already emitted 4-byte BE with three zero pad bytes).
- **Crash removal**: `cursorImage!`, `imagedataInt8!`, and `representations[0] as! NSBitmapImageRep` were all force-unwraps/casts on Cocoa-supplied data. Using `representations.first(where: { $0 is NSBitmapImageRep })` is the right fix — a cursor whose first rep is an `NSCGImageSnapshotRep` would previously have trapped.
- `cursorCoordinate` correctly guards `isFinite` before `UInt32(...)` — an unchecked `UInt32(nan)` traps in Swift, so the clamp test is guarding a real crash.
- `guard !cursorChangedCallbacks.isEmpty` early-returns *before* mutating `previousCursorImageHashes`, so no stale-state hazard from the added short-circuit. Nice detail.

## Issues / suggestions

**1. (minor, non-blocking) State advances before encoding can fail.**
`checkMouseCursor` sets `previousCursorImageHashes = cursorImageHashes` at line ~2392, but the later `guard let encoded = encodeCursorBitmap(...) else { return }` bails afterwards. If encoding fails once, that cursor is now "already seen" and will never be sent until the user switches to a different cursor and back. Previously this path crashed instead, so it's still an improvement — but consider resetting `previousCursorImageHashes = nil` (or moving the assignment after a successful encode) in the failure path. Same applies to the `continue` inside the hookAll loop.

**2. (question) Probe cadence is hardcoded and untested.**
`for delay in [16, 50]` is the actual behavioral fix for the reported bug, and it's the one thing with no test coverage. The two tests cover the mask and the hotspot math, both of which are the easy parts. Not blocking, but at minimum pull `[16, 50]` into a named constant with a comment on why those values — 16 ms is presumably one frame at 60 Hz; that intent is only implicit.

Also note each click now fires 6 `checkMouseCursor()` calls (down + 2 probes, up + 2 probes). `checkMouseCursor` runs SHA-256 over every bitmap rep, so a rapid click-heavy game does noticeably more hashing than before. Probably fine (cursors are tiny, and it's bounded), but worth being aware of; the `cursorChangedCallbacks.isEmpty` guard mitigates the no-listener case.

**3. (nit) `pointSize` fallback chain.**
```swift
let pointSize = bitmapRep.size.width > 0 && bitmapRep.size.height > 0 ? bitmapRep.size : image.size
```
If `image.size` is also zero, `cursorHotSpotInPixels` falls back to scale 1 — which for a 2x Retina cursor silently reports the hotspot at half its true pixel position rather than failing loudly. Acceptable degradation, just undocumented; a one-line comment would help.

**4. (nit) Dead code left behind.**
`previousCursorImage` is still assigned but appears unused elsewhere; and the stale commented-out `//print("default: \(cursorIndex!)")` now references a non-optional binding. Trivial cleanup while you're in here.

**5. `hookAll` bootstrap path** now requires `encodeCursorBitmap` to succeed for the *whole* `if let` chain, so a non-bitmap-rep cursor silently skips the initial cursor push (client sees no cursor until the first move). Previously it would have crashed, so again strictly better — but the client may want a fallback.

## Testing / security

- Test coverage is reasonable for the pure functions; `cursorHotSpotInPixels` and the mask are the two testable seams and both are covered. The 16/50 ms probe logic and `encodeCursorBitmap` byte layout are untested — an `encodeCursorBitmap` header-layout assertion would be cheap and would lock the wire format.
- No security concerns. `UInt32(clamping:)` and the finite guard remove the two integer-conversion trap sites; nothing reads attacker-controlled buffer lengths.

## Verdict rationale

Every finding is a nit or a pre-existing-behavior-improved-but-not-perfect edge case. The force-unwrap removal and the BE serialization fix are unambiguous wins, and the change is user-verified on the reported repro. Item 1 is the only one I'd genuinely like addressed, and it's a one-line follow-up rather than a merge blocker.

zhuhaichao518:
Addressed the review feedback in `ee20116`:

- retry unchanged cursor textures when bitmap encoding fails instead of marking them handled
- name and document the 16/50 ms transition probe schedule
- remove the unused previous cursor image state
- cover the probe schedule in XCTest

@hxyshadow review

hxyshadow (commented):
**💬 COMMENT(评论)**

**✅ APPROVE (re-review)**

## Code Review: #42 — fix(macos): refresh cursor textures and hotspot

**Verdict: Approved ✅** (0 issues, 0 open suggestions)
**Author:** @zhuhaichao518 | **Files:** 2 (+181 -88)
**Reviewed against:** `ee20116` (HEAD) — incremental over prior APPROVE at `a00f64e`

Re-review of the `ee20116` fixup. Verified each addressed item on disk at the new HEAD. All four land correctly; nothing regressed.

### Carry-forward check (prior review items)

- ✅ **Retry unchanged cursor textures when bitmap encoding fails** (was: state advanced before encode could fail → a one-time encode failure permanently marked the cursor "seen"). Fixed cleanly: `previousCursorImageHashes = cursorImageHashes` is now advanced only on the *success* tail of each branch.
  - Default/hookAll branch (`HardwareSimulatorPlugin.swift:2428`) gates the advance behind `updatedAllCallbacks`; a failed `encodeCursorBitmap` sets it `false` + `continue`, so state stays put and the next probe retries.
  - Cache branch (`:2469`) and fresh-image branch (`:2492`) advance only after `guard let encoded = encodeCursorBitmap(...) else { return }` passes.
  - The `previousCursorImage = cursorImage` line was also dropped, so no stale-image write survives the early return.
- ✅ **Name + document the 16/50 ms probe schedule** — `static let cursorTransitionProbeDelaysMs = [16, 50]` (`:2168`); the probe loop now reads `Self.cursorTransitionProbeDelaysMs` (`:2252`) with a comment noting one-frame/short-hitch coverage and that changes after 50 ms are picked up by the next mouse event. This is exactly the hoist + boundary-doc from my prior inline 💡.
- ✅ **Remove unused previous cursor image state** — `var previousCursorImage: NSImage?` declaration and its assignment removed; the stale `//print("default: \(cursorIndex!)")` comment (which referenced a now-non-optional binding) is gone too.
- ✅ **Cover the probe schedule in XCTest** — `RunnerTests.swift:45` asserts `cursorTransitionProbeDelaysMs == [16, 50]`, pinning the constant so a future edit to the cadence trips the test.

### ✅ Looks Good
- The retry semantics are the right trade-off: a truly-unencodable cursor now keeps retrying on subsequent mouse events instead of being silently swallowed forever, and the cost is bounded to actual user interaction (mouse events + the two probes, not a permanent timer) with the `cursorChangedCallbacks.isEmpty` early-out still short-circuiting the no-listener case.
- Retry gating is symmetric across all three branches — no branch advances state on a failed encode.

No new inline comments. Clean fixup — good to merge.

---

## Overview

The PR fixes two macOS cursor issues in `HardwareSimulatorPlugin.swift`:
1. **Missed cursor updates** — the global event monitor only watched move/drag events, so a cursor swapped in response to a click was missed. Now button down/up are in the mask, plus two bounded follow-up probes (16 ms / 50 ms).
2. **Wrong hotspot & inconsistent serialization** — three copy-pasted serialization blocks are collapsed into one `encodeCursorBitmap`, and hotspot scaling now derives from the bitmap rep instead of `NSScreen.backingScaleFactor`.

Also removes dead nil checks (`NSCursor`/`image` are non-optional) and stops boxing an optional `cursorIndex`.

## Strengths

- **Dedup is the real win.** Four near-identical byte-packing blocks → one function. The old blocks disagreed with each other (one applied `scaleFactor`, two didn't) — that inconsistency *was* the bug.
- **Removes latent traps.** `UInt8(cursorImage.representations[0].pixelsWide)` traps at runtime for width ≥ 256, and `UInt8(hotSpot.x)` traps on negative/NaN. `appendUInt32BE` + `cursorCoordinate`'s finite/clamp guard fix both. Wire format is unchanged for the previously-working range (`[0,0,0,w]` == BE u32).
- **Consistency guarantee**: hash, dimensions, and hotspot now all come from the *same* `bitmapRep`, instead of hashing one rep and measuring another.
- `previousCursorImageHashes` is now only committed after a successful send, so an encode failure doesn't permanently suppress that cursor.

## Issues & suggestions

**1. `pointSize` prefers `bitmapRep.size` over `image.size` — likely backwards** (`encodeCursorBitmap`)

```swift
let pointSize = bitmapRep.size.width > 0 && bitmapRep.size.height > 0
  ? bitmapRep.size
  : image.size
```

`NSCursor.hotSpot` is documented in the **image's** coordinate space (`image.size`, points). `NSBitmapImageRep.size` is only guaranteed to match that when it was explicitly set; a rep constructed from raw pixel data defaults to reporting its *pixel* dimensions. In that case `scaleX = pixelWidth / pixelWidth = 1` and the hotspot stays in points against a 2x bitmap — exactly the halved-hotspot bug this PR is fixing, just re-entering through a different door.

Suggest inverting the preference:
```swift
let pointSize = image.size.width > 0 && image.size.height > 0 ? image.size : bitmapRep.size
```
You verified against a real game cursor, so this may never trigger in practice — but the fallback ordering is worth a second look, and it's cheap to flip.

**2. Repeated-message risk in the default-cursor branch**

`updatedAllCallbacks` is set false when a hookAll callback fails to encode, so `previousCursorImageHashes` isn't committed. Correct for the failing callback — but the *non*-hookAll callbacks in the same loop already received their message, and will receive it again on every subsequent mouse event until encoding succeeds. Consider tracking per-callback delivery, or accepting the stale-hash cost only for the hookAll subset. Low severity (encode failure is rare), but it's an unbounded-message path.

**3. Probe timers are unconditional and uncoalesced**

Every button down/up schedules 2 main-queue blocks. A click is 2 events → 4 `checkMouseCursor()` calls in 50 ms, each doing a SHA-256 over all bitmap reps. For rapid clicking (games — the exact target scenario) this is a steady stream of hashing on the main thread. `checkMouseCursor` now early-returns when there are no callbacks, which helps, but consider coalescing: cancel/skip a pending probe if one is already scheduled (a simple `pendingProbeCount` or `DispatchWorkItem` you cancel and re-arm).

**4. Test coverage gaps**

The added tests cover the mask and `cursorHotSpotInPixels` — good, and the NaN/negative clamp test is the right instinct. But nothing covers `encodeCursorBitmap`, which is the function that actually changed the wire format. A test that builds a synthetic `NSImage` with a known 2x `NSBitmapImageRep` and asserts the 21-byte header (tag, width, height, hotspot, hash) would lock in the format and would have caught issue #1 directly.

`testCursorMonitorIncludesButtonTransitions` asserting `cursorTransitionProbeDelaysMs == [16, 50]` is a change-detector test — it fails on any tuning without indicating breakage. Consider asserting invariants instead (non-empty, sorted, all ≤ some bound).

## Style / minor

- `cursorTransitionProbeDelaysMs` would be better as `[Int]` with an explicit type annotation and, ideally, a `Duration`/`DispatchTimeInterval` name that carries the unit at the use site rather than in the identifier suffix.
- Indentation of the new `guard` bodies (2-space continuation) differs from the surrounding 4-space blocks; the file is already mixed, so not a blocker.
- Nice cleanup dropping the stale `// TODO: 为什么获取到的非系统自带图片的hotSpot不对？暂缓方案` — this PR is the actual answer to that TODO.

## Risk

Low-to-moderate. Wire format is backward compatible for cursors ≤ 255 px with in-range hotspots, and strictly more correct outside that range. The main risk is the `pointSize` preference (#1) regressing Retina hotspots for cursor images whose rep size reports pixels; user verification on the reported failing case mitigates this but doesn't cover the whole space.

Overall this is a solid, well-scoped fix that removes more code than it adds and eliminates two runtime-trap paths. Worth addressing #1 before merge if you can retest on a Retina display with a non-default cursor.

hxyshadow (commented):
**💬 COMMENT(评论)**

## Overview

Refactors macOS cursor handling in `HardwareSimulatorPlugin.swift`:
- Extracts the four duplicated cursor-serialization blocks into one `encodeCursorBitmap(image:hotSpot:)`.
- Adds mouse down/up to the cursor monitor mask plus two bounded follow-up probes (16 ms, 50 ms) to catch game cursor swaps.
- Replaces the `backingScaleFactor` + "≤32px ⇒ scale 1.0" heuristic with a real pixel/point ratio for hotspot conversion.
- Drops dead nil-checks on non-optional `NSCursor` values, and adds 4 XCTests.

## Correctness

- **Wire-format widening is a real (good) fix, but a compat point.** Old code emitted `[0,0,0,UInt8(pixelsWide)]` — anything ≥256 px silently wrapped, and `UInt8(Int(hotSpot.x * scaleFactor))` would *trap* on out-of-range values. The new `appendUInt32BE` + `UInt32(clamping:)` fixes both a truncation and a crash path. Worth confirming the Dart/receiver side already parses these as big-endian u32 (it should, given the byte layout is unchanged) — if any consumer read only the low byte, this changes behavior for large cursors.
- **`previousCursorImageHashes` update moved after successful send** in all three branches, with `updatedAllCallbacks` guarding the default-cursor hookAll path. This is a correctness improvement: a failed encode no longer poisons the cache and permanently suppresses that cursor. Nice detail.
- **Early `guard !cursorChangedCallbacks.isEmpty`** is correct and skips the expensive sha256-over-all-reps when nobody is listening.
- `representations.first(where: { $0 is NSBitmapImageRep })` vs. the old `representations[0] as! NSBitmapImageRep` — removes a force-cast crash for cursors whose first rep isn't a bitmap. Good. Minor: for multi-rep (1x/2x) cursors this may pick the 1x rep; since the hotspot scale is derived from *that same rep*, the result stays self-consistent, so no bug — just lower fidelity than picking the largest rep.

## Suggestions (non-blocking)

- `handleCursorMonitorEvent` also calls `sendCursorPositionToAll()` on every down/up now (previously move/drag only). Harmless, but it's an unannounced increase in position-message traffic; worth a line in the commit message.
- The two probes call `checkMouseCursor()` unconditionally, and each one does a sha256 over every bitmap rep. Per click that's 3 hashes instead of 1. Fine at click frequency, but if rapid-fire clicking in a game becomes a hotspot, consider coalescing (drop a probe if another `checkMouseCursor` already ran within the window).
- Probes are scheduled with `[weak self]` on the main queue — good — but they aren't cancelled when the last callback unregisters. The `cursorChangedCallbacks.isEmpty` guard makes them no-ops, so this is fine; just noting it's guarded by that guard, not by cancellation.
- `cursorCoordinate` uses `value > 0`, so exactly-0 falls to the `return 0` path — same result, but `value >= 0` would read more obviously as "clamp negatives".
- The comment on `pointSize` fallback (image.size vs bitmapRep.size) is genuinely useful — good that it explains *why*, not *what*.

## Tests

Solid for a plugin layer that's usually untested: mask contents, probe delays, hotspot scaling, invalid-coordinate clamping, and a full round-trip byte-layout assertion (`21 + 64*48*4`, header fields at fixed offsets). The byte-layout test in particular locks in the wire format, which is the riskiest part of this change.

Gaps, if you want to extend later: no test for the `encodeCursorBitmap` → `nil` path (non-bitmap rep) or for the `updatedAllCallbacks == false` cache-not-poisoned behavior, which is the subtlest logic added here.

## Security / risk

Nothing security-relevant. Main residual risk is the receiver-side parsing of the widened dimension/hotspot fields; PR body reports user-verification in CloudPlayPlus, which covers that.

Net: a clean deduplication that removes two crash paths and a truncation bug, with tests pinning the format. Already merged; no blockers.

